PR #53, opened 2023-05-30
Prepare majo release

Adam-it:
## 🎯 Aim

The aim is to prepare for major release.


Adam-it:
- [ ] double check local package before merge

Issue #212, opened 2024-04-10
🐞 [Bug]: Serve fails if folder path has space


Tanddant:
## 📝 Describe the bug

When I press the serve button and the path to the folder has a space in it PowerShell `cd` fails

i.e if my project is stored at:

`d:\Client name with spaces\Project 1`

I'll get the following output:

``` PowerShell
cd d:\Client name with spaces\Project 1
> Set-Location: A positional parameter cannot be found that accepts argument 'name'.
```

## 👣 Steps To Reproduce

1. Open a project that's stored in a path with a space in it
2. Press the serve button

## 📜 Expected behavior

Serve my project.

## 💻 Desktop (please complete the following information):

 - OS: Windows 11
 - Browser N/A
 - Version 3.0.0

## 🤔 Additional context

Seems like all that's needed is to wrap the paramater containing the path in a set of quotes, I would love to help, but I have idea where to look in the code, so if anyone can point me in a direction I would love to help fix it

Adam-it:
hi @Tanddant 👋
Sorry to hear you are having trouble using SPFx Toolkit 😔 and thank you so so much for reporting the bug 🐞.
From what you descried I also (as you suggested) think that the path should be just wrapped in single quotes 👍. 
Seems I totally forgot one may use `space` in folder names 😅.

... anyway. like 90% this is the place which should be fixed
https://github.com/pnp/vscode-viva/blob/c11c78f5ebbbd4965c3943b6287a15cc6f7958ca/src/services/TerminalCommandExecuter.ts#L120

The `TerminalCommandExecuter.runCommand` method executes every of the gulp task commands. At the beginning it sets the user newly created terminal in right position which needs to be fixed.

Would you like to give it a try?
I am still working on better contributing guidance on this repo as it is a bit problematic but you may always follow along with the current [Contribution guidelines](https://github.com/pnp/vscode-viva/blob/main/contributing.md) and if anything seems wrong or stopping you from work just let me know either here or directly 👍

Tanddant:
For sure I'll give it a go! - never worked on a VS Code extension before, new stuff is always fun, feel free to assign me this task 😊

Adam-it:
> For sure I'll give it a go! - never worked on a VS Code extension before, new stuff is always fun, feel free to assign me this task 😊

Sold 🔨
VS Code extensions it's a world of its own 😅. Just like SPFx 😜 
Let me know if you will need any help.

Adam-it:
[...]

PR #214, opened 2024-04-13
🐛 - Fixed bug when there's a space in the folder path causing serve t…

Tanddant:
…o fail

## 🎯 Aim

Goal is to solve #212 

## 📷 Result

Folder path now wrapped in quotes
image

## ✅ What was done

Hopefully nothing else needs to be done 🙌

- [] not done
- [X] done

## 🔗 Related issue

Closes: #212

### DIFF
--- a/src/services/TerminalCommandExecuter.ts
+++ b/src/services/TerminalCommandExecuter.ts
@@ -117,7 +117,7 @@ export class TerminalCommandExecuter {
         currentProjectPath = join(currentProjectPath, 'src');
       }
 
-      TerminalCommandExecuter.runInTerminal(`cd ${currentProjectPath}`, terminal);
+      TerminalCommandExecuter.runInTerminal(`cd "${currentProjectPath}"`, terminal);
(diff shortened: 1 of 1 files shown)


Adam-it:
@Tanddant thank you for your awesome work 👏. I will try to review it ASAP

Adam-it:
@Tanddant I marked this as draft as in as we should target all PR's to `dev` branch first and not to `main` and also we should start from `dev` (don't worry about this. I am totally aware the contribution guidance is really minimalistic and I want to work on it this week). Since this branch started from `main` it now has some changes that should not be here but I will resolve this by rebasing `dev` against `main` on this repo, no problem. So in order not to merge this by mistake I set this to draft for that time, just so that you are aware and no action is required from your side 👍.

Adam-it:
rebased ✅

Adam-it (approved):
checked locally ✅
image

Works like a charm 🪄
👏👏👏

Issue #278, opened 2024-07-25
🐞 Bug report: Account pane not populated (no username, sites list, service health)


germain-italic:
### ⭐ Priority

(Medium)😢 I'm annoyed but I'll live

### 📝 Describe the bug

- I can sign in and sign out on my tenant as full admin
- I can see the Tenant App Catalog
- **But** there is no Site App Catalogs menu
- I can deploy sppkg using the extension
- **But** I cannot select in which catalog (therefore my app is deployed tenant-wide)

### 👣 Steps To Reproduce

- Install extension
- Check installed dependencies
- Create a new project
- Sign in
- Wait and nothing appears under Account
- Try to use Refresh Account View, same result
- Using `m365 spo site list` works as expected

### 📜 Expected behavior

- Account name visible
- List of sites in the tenant visible

### 📷 Screenshots

![image](https://github.com/user-attachments/assets/4cfea96d-1ed8-4397-8521-07a37d76167a)

### ❓SharePoint Framework Toolkit version

v3.4.0

### ❓Node.js version

v18.17.1

### 🤔 Additional context

From the Output window:

```
> M365 CLI - status: "Logged out"
[2024-07-25T22:34:24.121Z] [INFO]
> Device code: DJXNFPDQR
[2024-07-25T22:34:36.890Z] [ERROR]
> M365 CLI - login: [object Object]
[2024-07-25T22:34:36.891Z] [INFO]
> M365 CLI - Logged in to M365
[2024-07-25T22:34:36.894Z] [INFO]
> M365 CLI - status: {
  "connectedAs": "me@mydomain.com",
  "authType": "DeviceCode",
  "appId": "31359c7f-bd7e-475c-86db-fdb8c937548e",
  "appTenant": "common",
  "cloudType": "Public"
}
[2024-07-25T22:34:36.906Z] [INFO]
> Running CLI command: spo tenant appcatalogurl get
"https://mytenant.sharepoint.com/sites/catalog_app"
[2024-07-25T22:34:43.154Z] [INFO]
> Running CLI command: spo site appcatalog list
[]
[2024-07-25T22:34:44.154Z] [INFO]
> Running CLI command: tenant serviceannouncement health list
```

Adam-it:
hey @germain-italic thanks for opening up the issue and letting us know 👍 You Rock 🤩.
From the log you provided it seems like there is already something wrong with login to CLI for M365 
[...]

PR #284, opened 2024-07-28
Improves account and env view.  Closes: #278

Adam-it:
## 🎯 Aim

The aim is to make account and enviroment views more bulletproof and properly display views when CLI command extecution gets rejected due to lack of permission

## 📷 Result

image

## ✅ What was done

- [X] Added simple error handling to cover the reject case

## 🔗 Related issue

Closes: #278


src/services/CliActions.ts
```diff
@@ -51,21 +51,25 @@ export class CliActions {
...
+        const siteAppCatalogsJson: SiteAppCatalog[] = JSON.parse(siteAppCatalogs);
+        siteAppCatalogsJson.forEach((siteAppCatalog) => appCatalogUrls.push(`${siteAppCatalog.AbsoluteUrl}/AppCatalog`));
+      }
 
-    EnvironmentInformation.appCatalogUrls = appCatalogUrls ? appCatalogUrls : undefined;
-    return EnvironmentInformation.appCatalogUrls;
+      EnvironmentInformation.appCatalogUrls = appCatalogUrls ? appCatalogUrls : undefined;
+      return EnvironmentInformation.appCatalogUrls;
+    } catch {
+      return undefined;
```
Adam-it, line 71: now that I think about it, thats not that good that try catch is now part of logic but this is a bit of a workaround when user has some scopes not granted for CLI app

Adam-it:
ready to merge 🚀

Issue #285, opened 2024-07-29
💡 [Feature]: Add setting to enable/disable the .nvmrc file creation


GuidoZam:
### 🎯 Aim of the feature

Add a new additional step setting when scaffolding a new project, the additional `YES/NO` switch will have a label 'Create node version manager configuration file (.nvmrc)'. This setting will be added after the already existing ones:

![image](https://github.com/user-attachments/assets/71a58345-0113-4170-a701-d4eee292612e)

The setting, if set to `YES`, will create the `.nvmrc` file when scaffolding the new project.

Other than adding the new additional step two new extension settings will be added:
- `createNodeVersionFileDefaultValue`: this setting will define what's the default value for the additional scaffold step. By default the value will be `NO`.
- `nodeVersionManagerFile`: this setting will define what Node version configuration file to use. If the user has selected NVM as Node versioning software this setting won't display, instead if the NVS versioning software is used the user can choose the configuration file to use, the available choices are:
  - `.nvmrc`
  - `.node-version`

### 📷 Images (if possible) with expected result

_No additional remarks_

### 🤔 Additional remarks or comments

_No additional remarks_

Adam-it:
@GuidoZam thanks for this suggestion 👍. I think it is a really awesome idea, let's give it though and discuss different options and then prototype(draft/spec) how it should be done before we open it up. 
I will try to get back with more comments and questions regarding this issue later during the day 👍

Adam-it:
@GuidoZam some feedback from my side on this idea:
1. This new setting seems strictly related to project scaffolding. VS Code extension setting are usually something most of us (developers)  do not change 😉 so if we put this there it will get little visibility. Besides that we already have some additional 'settings' that one may be interested when scaffolding new project and this may be found in the 3rd step of the scaffolding form. IMO I would rather add this as a new option in the 'Additional Steps' section in the scaffolding form. Just another `YES/NO` switch with label like 'create node version manager configuration file (.nvmrc)' or something similar.
![image](https://github.com/user-attachments/assets/5bfc167d-8a4e-4e6c-87f1-3670ada17872)
2. Then if the above suggestion would be fine with you in the VS code extension settings I would rather add a setting if this should be 'turned on' (so set as YES) by default. Otherwise by default it will be NO which should be this default... default value 😅
[...]

PR #315, opened 2024-10-05
Nvmrc new settings

GuidoZam:
## 🎯 Aim

Added a new additional step setting when scaffolding a new project to specify if the node version manager configuration file should be created or not.

## 📷 Result

image

image

## ✅ What was done

- [X] Added a new project additional step to create or not the node version manager configuration file.
- [X] Added new VSCode setting `createNodeVersionFileDefaultValue` to define the default value of the new project additional step.
- [X] Added new VSCode setting `nodeVersionManagerFile` to specify which configuration file to be created.

## 🔗 Related issue

Closes: #285


GuidoZam:
Removed as there are leftovers commits, replacing it with an up to date PR.

PR #316, opened 2024-10-05
Added additional settings.

GuidoZam:
## 🎯 Aim

Added a new additional step setting when scaffolding a new project to specify if the node version manager configuration file should be created or not.

## 📷 Result

image

image

## ✅ What was done

- [X] Added a new project additional step to create or not the node version manager configuration file.
- [X] Added new VSCode setting `createNodeVersionFileDefaultValue` to define the default value of the new project additional step.
- [X] Added new VSCode setting `nodeVersionManagerFile` to specify which configuration file to be created.

## 🔗 Related issue

Closes: #285


Adam-it:
Awesome 🤩😍. I will review this ASAP

src/services/actions/Scaffolder.ts
```diff
@@ -147,6 +147,45 @@ export class Scaffolder {
...
+			WebviewCommand.toWebview.createNodeVersionManagerFile,
+			value
+		);
+	}
+
+	/**
+	 * Returns the value of the nodeVersionManager setting and sends it to the webview.
+	 */
+	public static async nodeVersionManager() {
+		const value = getExtensionSettings<string>('nodeVersionManager', false);
```
Adam-it, line 181: `false` should not be a default value of a setting which we rather expect to return a string. I also introduced a bug fix connected to that just a while ago which will most probably make this not working in order to ensure correct types. 
lets update  this to `nvm` as default
```suggestion
		const value = getExtensionSettings<string>('nodeVersionManager', 'nvm');
```

src/services/actions/Scaffolder.ts
```diff
@@ -147,6 +147,45 @@ export class Scaffolder {
...
+			WebviewCommand.toWebview.createNodeVersionFileDefaultValue,
+			value
+		);
+	}
+
+	/**
+	 * Returns the value of the nodeVersionManagerFile setting and sends it to the webview.
+	 */
+	public static async nodeVersionManagerFile() {
+		const value = getExtensionSettings<string>('nodeVersionManagerFile', false);
```
Adam-it, line 169: also here the default value should be string. Since in `package.json` we specify the default is `.nvmrc` lets do the same here

src/services/actions/Scaffolder.ts
```diff
@@ -225,6 +264,29 @@ export class Scaffolder {
...
+                    content += ` ${ProjectFileContent.createNodeVersionFile}`;
+                    break;
+                }
+                break;
+                case 'none':
+                // If the node version manager is none, do not create any file
+                break;
+							}
+						}
+
```
Adam-it, line 289: I just must ask since GH is rendering the whitespace very strange in this PR. 
what is encoding and set and space indentation in VS Code?
image

src/services/actions/Scaffolder.ts
```diff
@@ -225,6 +264,29 @@ export class Scaffolder {
...
+                switch (newSolutionInput.nodeVersionManagerFile) {
+                  case '.nvmrc':
+                    content += ` ${ProjectFileContent.createNVMRCFile}`;
+                    break;
+                  case '.node-version':
+                    content += ` ${ProjectFileContent.createNodeVersionFile}`;
+                    break;
+                }
+                break;
+                case 'none':
```
Adam-it, line 284: perfect case to be default as well. Lets add this case as `default` so that if in future we will add some new node version manager by default we will not create any file 

```suggestion
                case 'none':
                default:
```

src/constants/WebviewCommand.ts
```diff
@@ -17,5 +22,10 @@ export const WebviewCommand = {
     validateComponentName: 'validate-component-name',
     addSpfxComponent: 'add-spfx-component',
     createAppReg: 'create-app-reg',
+		createNodeVersionFileDefaultValue: 'should-create-node-version-file',
+		nodeVersionManager: 'node-version-manager',
+		nodeVersionManagerFile: 'node-version-manager-file',
+		createNVMRCFile: 'create-nvmrc-file',
+		createNodeVersionManagerFile: 'node-version-manager-file',
```
Adam-it, line 29: seems those two are never sent from webview to vscode. Lets remove them

```suggestion
```

src/constants/WebviewCommand.ts
```diff
@@ -5,6 +5,11 @@ export const WebviewCommand = {
     folderPath: 'folder-path',
     validateSolutionName: 'validate-solution-name',
     validateComponentName: 'validate-component-name',
+		createNodeVersionFileDefaultValue: 'should-create-node-version-file',
+		nodeVersionManager: 'node-version-manager',
+		nodeVersionManagerFile: 'node-version-manager-file',
+		createNVMRCFile: 'create-nvmrc-file',
```
Adam-it, line 11: is this used? I am almost sure this one is not used and we may remove it. We use a similar constant in `ProjectFileContent.ts`

Adam-it (changes requested):
@GuidoZam Awesome work so far 👏👏👏👏👏. You Rock 🤩
I left a few comments. Please do give them a recheck.
When testing I also found some 2 strange behaviors we should fixup before we merge:

I don't know why but when I select creating nvm file together with other dependencies 
image
it is set in the project.pnp file 
image
but as a result in terminal I only get the nvm file installed no other npm install is running 
image
I don't know why but this seem to be only like this when I have this new dependency set otherwise it works normally and installed everything so it seems like a bug

also 
when you reset the create nvm file to default it will be set to true not false by default
image

most probably this is due to the fact that in `package.json` we set this a string "false" not `false`

Adam-it:
@GuidoZam I also forgot to mention that we should update the readme part with additional step in scaffolding form to now include this new option as well together with screenshots

src/services/actions/Scaffolder.ts
GuidoZam: It's strange but in some way VS Code has added tabs instead of spaces to indent, I'll fix this as well.

data/sp-dev-fx-samples.json
Adam-it, line 1: @GuidoZam just a quick heads up. This file should not be modified during a PR . 
in the repo there is a separate GH Action bot that modifies this one

Adam-it (changes requested):

GuidoZam:
Hi @Adam-it, I verified and fixed all your reviews.
The two bugs are resolved:
- The first one I think was a different behavior produced by the custom name of the terminal. I've removed that customization and everything's seems to work fine.
- About the second bug you were right, it was just a type mistake, updating the `package.json` type resolved the issue.

I've also updated the README file, have a look and tell me if this way is ok for you. There are some new images and text for the Node version manager support (section 12 of the README).

About the `data/sp-dev-fx-samples.json` file I didn't update that on purpose, I guess I've missed something while mergin' the branches.

Let me know if there's something else to update (hopefully not 🤞).

Adam-it (approved):
LGTM  👍 and works perfectly 👏
I did a couple of small fixups

PR #319, opened 2024-10-06
Present solutions deployed per app catalog. Closes #215

### LINKED ISSUE #215: 💡 [Feature]: Present solutions deployed per app catalog
### 🎯 Aim of the feature

The aim of this issue is to show apps deployed to each app catalog (similar way like we now present tenant-wide extensions for tenant level app catalog)

Additionally, we could introduce the following features:
- allow to disable and enable a solution
- allow to move solutions between app catalogs (from site level to tenant level and the other way around)
- allow to remove and retract a solution
- clicking on app link should go to the solution appStore.aspx page
- allow to add a webpart to all sites

we should add an extension setting that will allow to hide this information in the UI

image

parent issue: #15

Saurabh7019:
## 🎯 Aim

> Present solutions deployed per app catalog and add additional app-related actions (deploy, retract, remove, enable, disable etc.)

## 📷 Result

image

## ✅ What was done

- [X] Dynamic loading of children for ActionTreeItem
- [X] Present solutions deployed per app catalog
- [X] Allow to disable and enable a solution
- [X] Allow to deploy and retract a solution
- [X] Allow to remove a solution
- [X] Clicking on app link should go to the solution appStore.aspx page
- [] Allow to move solutions between app catalogs (from site level to tenant level and the other way around)
- [] Allow to add a webpart to all sites

## 🔗 Related issue

> Closes #215


src/panels/CommandPanel.ts
```diff
@@ -169,37 +168,110 @@ export class CommandPanel {
...
+        catalogItems.push(tenantWideExtensionsNode);
       }
 
+      const tenantAppCatalogNode = new ActionTreeItem(tenantAppCatalogUrl.replace(origin, '...'), '', { name: 'globe', custom: false }, TreeItemCollapsibleState.Collapsed, undefined, undefined, 'sp-app-catalog-url', undefined,
+        async () => {
+          const tenantAppCatalogApps = await CliActions.getAppCatalogApps();
+          const tenantAppCatalogAppsList: ActionTreeItem[] = [];
+
+          if (tenantAppCatalogApps && tenantAppCatalogApps.length > 0) {
+            const appStoreUrl = `${tenantAppCatalogUrl}/_layouts/15/appStore.aspx`;
```
Adam-it, line 206: currently when clicking on any app we will be redirected to the general appStore link. The idea was to be redirected to the `/appDetail/` page  to see details of that app from the store 
image

src/services/actions/CliActions.ts
```diff
@@ -15,8 +15,16 @@ import { PnPWebview } from '../../webview/PnPWebview';
 import { parseYoRc } from '../../utils/parseYoRc';
 import { CertificateActions } from './CertificateActions';
 import path = require('path');
+import { ActionTreeItem } from '../../providers/ActionTreeDataProvider';
 
 
+interface AppCatalogApp {
```
Adam-it, line 21: lets move this to `model` folder

Adam-it (changes requested):
@Saurabh7019 awesome work so far 👏👏👏 You Rock 🤩
I added few minor comments. Please do give them a recheck 🙏
Also I noticed some things when testing we might want to fixup before we merge:
It seems I was always unable to remove an app 🤔
image

also it seems that now it is impossible to redirect to a tenant app catalog or site-level app catalog when clicking on the link. This is due to the fact that those are now groups not single items. But still I think it was pretty nice to go to the app catalog by clicking on the item. What if we add this as a additional action next to the group? when clicking on it we could still open the appcatalog to get easy transfer to it. What do you think?

This PR is really awesome and I already see a lot of potential it opens up to more amazing functionalities this area may bring 🤩

Saurabh7019:
> But still I think it was pretty nice to go to the app catalog by clicking on the item

Oh, I missed that! I'll include it today. Should I just make the node itself a link, or would you prefer an action for it? The App Catalog node already looks like a link, so as a user I would think it should be clickable. Users can click on the node to navigate directly to the App Catalog, and click the arrow on the left to expand and see the list of apps.

image

or, I can provide additional action, if you prefer that?

Adam-it:
@Saurabh7019 I'll leave it up to you. For me any of the options you mentioned are fine 👍 as long as we keep this functionality as I think it is just helpful. 
I will leave the decision up to you 👍 this is an open-source and community-driven tool and I am confident you will pick the best option

Adam-it:
@Saurabh7019 I also not just realized we should update the guidance in the readme and screenshots that present this part of toolkit to keep this up to date 👍

Adam-it:
Awesome work 👏👏👏. You Rock 🤩

PR #323, opened 2024-10-13
Added tour response for Upgrade and Validate actions, Closes #188

### LINKED ISSUE #188: 💡 [Feature]: The `Upgrade` and `Validate` actions could return also code tour response
## 🎯 Aim of the feature

As internally those actions use the `project upgrade` and `project doctor` CLI for M365 commands and those support also the `tour` output mode which create a code tour for the current project with the findings. We could extend those functionalities so that we return the markdown sum up report and the tour in one go.

## Proposed implementation

In the `CliActions` service we the upgrade and validate methods. Each of those just run the CLI command under the hood with `md` output mode.
image

Depending on the what user have saved in the setting (see below) we should execute either the command with `md` or `tour` output or execute the commands twice with both output for both

## Additional Proposal

Similar to node settings we could add additional setting for those features. For each of those features: `upgrade` and `validate` we could provide a dropdown with options that define what report type should be generated. Either: `markdown` (as we have now), `code tour`, `both` (which should be the default value)
image

nicodecleyre:
## 🎯 Aim

Adding the tour options and both md & tour options as settings for the upgrade and validate actions

## 📷 Result

image

## ✅ What was done

- [X] Added `Md`, `Tour`, `Both` as upgrade options in the settings 
- [X] Added `Md`, `Tour`, `Both` as validate options in the settings 
- [X] Added tour logic to code for the upgrade option
- [X] Added tour logic to code for the validate option

## 🔗 Related issue

Closes: #188

### DIFF
--- a/src/services/actions/CliActions.ts
+++ b/src/services/actions/CliActions.ts
@@ -505,17 +505,13 @@ export class CliActions {
         const projectUpgradeOutputMode: string = getExtensionSettings('projectUpgradeOutputMode', 'both');
         let resultMd: any;
 
-        if (projectUpgradeOutputMode === 'markdown') {
+        if (projectUpgradeOutputMode === 'markdown' || projectUpgradeOutputMode === 'both') {
           resultMd = await CliExecuter.execute('spfx project upgrade', 'md');
           CliActions.handleMarkdownResult(resultMd, wsFolder, 'upgrade');
-        } else if (projectUpgradeOutputMode === 'code tour') {
+        } 
+        
+        if (projectUpgradeOutputMode === 'code tour' || projectUpgradeOutputMode === 'both') {
           await CliExecuter.execute('spfx project upgrade', 'tour');
-          CliActions.handleTourResult(wsFolder, 'upgrade');
-        } else {
-          resultMd = await CliExecuter.execute('spfx project upgrade', 'md');
-          await CliExecuter.execute('spfx project upgrade', 'tour');
-
-          CliActions.handleMarkdownResult(resultMd, wsFolder, 'upgrade');
           CliActions.handleTourResult(wsFolder, 'upgrade');
         }
       } catch (e: any) {
@@ -844,11 +840,15 @@ export class CliActions {
     }
     
     const tourFilePath = path.join(wsFolder.uri.fsPath, '.tours', `${fileName}.tour`);
-    
+    await workspace.fs.stat(Uri.file(tourFilePath));
+
+    // A timeout is needed so Codetour can find the tour file
     if (fs.existsSync(tourFilePath)) {
-      await commands.executeCommand('codetour.startTour');
+      setTimeout(() => {
+        commands.executeCommand('codetour.startTour');
+      }, 500);
     } else {
-      Notifications.error('Validation tour file not found. Cannot start Code Tour.');
+      Notifications.error(`${fileName}.tour file not found in path ${path.join(wsFolder.uri.fsPath, '.tours')}. Cannot start Code Tour.`);
     }
   }
 }
--- a/src/services/actions/CliActions.ts
+++ b/src/services/actions/CliActions.ts
@@ -1,6 +1,6 @@
 import { readFileSync, writeFileSync } from 'fs';
 import { Folders } from '../check/Folders';
-import { commands, Progress, ProgressLocation, Uri, window, workspace } from 'vscode';
+import { commands, Progress, ProgressLocation, Uri, window, workspace, WorkspaceFolder } from 'vscode';
 import { Commands, ContextKeys, WebViewType, WebviewCommand, WorkflowType } from '../../constants';
 import { AppCatalogApp, GenerateWorkflowCommandInput, SiteAppCatalog, SolutionAddResult, Subscription } from '../../models';
 import { Extension } from '../dataType/Extension';
@@ -17,7 +17,8 @@ import { parseCliCommand } from '../../utils/parseCliCommand';
 import { CertificateActions } from './CertificateActions';
 import path = require('path');
 import { ActionTreeItem } from '../../providers/ActionTreeDataProvider';
-
+import { getExtensionSettings } from '../../utils/getExtensionSettings';
+import * as fs from 'fs';
 
 export class CliActions {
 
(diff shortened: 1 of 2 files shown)


Adam-it:
Awesome. Thank you for another awesome contribution. I will review it ASAP

src/services/actions/CliActions.ts
```diff
@@ -452,21 +454,22 @@ export class CliActions {
...
-          Notifications.error(result.stderr);
+        const projectValidateOutputMode: string = getExtensionSettings('projectValidateOutputMode', 'both');
+        let resultMd: any;
+
+        if (projectValidateOutputMode === 'markdown') {
+          resultMd = await CliExecuter.execute('spfx project doctor', 'md');
+          CliActions.handleMarkdownResult(resultMd,wsFolder);
+        } else if (projectValidateOutputMode === 'code tour') {
+          await CliExecuter.execute('spfx project doctor', 'tour');
+          commands.executeCommand('codetour.startTour');
```
Adam-it, line 465: the problem with this approach is that it will run code tour even if nothing was found. in this case we see a picker to run tour but there is non to run. this may be very confusing and I suggest we should run the code tour command only when something was created, so the CLI command did return something
image

The same applies to the Upgrade report

Adam-it (changes requested):
@nicodecleyre Awesome work so far 👏👏👏👏
I tested it and works really good. During my tests and review I noticed some small details we could sort out before we merge 👍
Please do give them a double check 🙏

Adam-it:
@nicodecleyre I also did some merging to the dev branch. May I kindly ask you to rebase against latest dev

Adam-it:
@nicodecleyre we just rebased `dev` branch with latest `main`. 
May I kindly ask you to align your branch 🙏 Cheers

src/services/actions/CliActions.ts
```diff
@@ -15,6 +15,8 @@ import { PnPWebview } from '../../webview/PnPWebview';
 import { parseYoRc } from '../../utils/parseYoRc';
 import { CertificateActions } from './CertificateActions';
 import path = require('path');
+import * as fs from 'fs';
```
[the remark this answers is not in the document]
nicodecleyre: In the original commit no, but now we do as we have to check if the file exists

src/services/actions/CliActions.ts
nicodecleyre: `spfx project upgrade` & `spfx project doctor` don't return anything if the option is `tour`, so we have to check if the file exists

nicodecleyre:
> @nicodecleyre we just rebased `dev` branch with latest `main`. May I kindly ask you to align your branch 🙏 Cheers

Still working on it, having a little trouble with the rebase 😆

src/services/actions/CliActions.ts
```diff
@@ -479,21 +480,21 @@ export class CliActions {
...
+
+        if (projectUpgradeOutputMode === 'markdown') {
+          resultMd = await CliExecuter.execute('spfx project upgrade', 'md');
+          CliActions.handleMarkdownResult(resultMd, wsFolder, 'upgrade');
+        } else if (projectUpgradeOutputMode === 'code tour') {
+          await CliExecuter.execute('spfx project upgrade', 'tour');
+          CliActions.handleTourResult(wsFolder, 'upgrade');
+        } else {
+          resultMd = await CliExecuter.execute('spfx project upgrade', 'md');
+          await CliExecuter.execute('spfx project upgrade', 'tour');
```
Adam-it, line 494: I was wondering if we could just have `if` blocks in which we check for either the specific type/method, like `markdown` or `tour`  and check for `both`. In this case we don't need to check explicitly for `both` and we will avoid code repetition. What do you think? I guess it should work

```suggestion
        if (projectUpgradeOutputMode === 'markdown' || projectUpgradeOutputMode === 'both') {
          resultMd = await CliExecuter.execute('spfx project upgrade', 'md');
          CliActions.handleMarkdownResult(resultMd, wsFolder, 'upgrade');
        } 
        
        if (projectUpgradeOutputMode === 'code tour' || projectUpgradeOutputMode === 'both') {
          await CliExecuter.execute('spfx project upgrade', 'tour');
          CliActions.handleTourResult(wsFolder, 'upgrade');
        }```

File: src/services/actions/CliActions.ts [hunk 4]
```diff
@@ -788,4 +790,43 @@ export class CliActions {
...
+      Notifications.error('Workspace folder is undefined. Cannot start Code Tour.');
+      return;
+    }
+    
+    const tourFilePath = path.join(wsFolder.uri.fsPath, '.tours', `${fileName}.tour`);
+    
+    if (fs.existsSync(tourFilePath)) {
+      await commands.executeCommand('codetour.startTour');
+    } else {
+      Notifications.error('Validation tour file not found. Cannot start Code Tour.');
```
Adam-it, line 829: the error may be misleading as it may not always be the validation tour but also it may be the upgrade tour

Adam-it (changes requested):
@nicodecleyre I think we are in a much better state now. I played with it a lot and I think it is a nice feature. I noticed some two small improvements we could double check before we proceed but those are not that important I guess.

What is my main concern, and this is more a question if you experienced the same behavior, is:
when I had `both` for validation and upgrade I noticed that sometimes when clicked on the upgrade action even though the tour was generated and the json was present te codeTour extension did not start it but instead it asked which tour should it run and none was available? I was wondering if the tour extension triggered just before the tour json was present 🤔 Maybe  we could add a bit of wait as I checked this is already awaited so I am not sure  why this behavior and I would need a bit deeper dive on this myself
image

And then the strange part. When I had the validation tour created first and then I clicked on the upgrade action the extension kicked of the validation tour 😮 instead of asking which tour should be triggered?
image

I was wondering did you maybe researched if there was a way to pass down to codeTour which specific tour should start, so we could avoid the picker when we have both the validation and upgrade tours? I will try to research this myself as well.

Other than the above concerns/questions I think we are all good and getting closer to get this merged 👍👍I wanted to make this feature perfect before we proceed 🤩. I will try to recheck some of the stuff mentioned above myself next week

nicodecleyre:
Hi Adam, you're right, there seems to be a sort of hickup. It seems that the setup right now "still lives in the past". I did some testing and this are the results:
- run validate: no code tour selectable
- run validate: validate code tour selectable
- run upgrade: validate code tour selectable
- run validate: validate & upgrade code tour selectable.

So it seems that only the tour from the previous run is selectable. Timeout doesn't seem to solve the problem right away. Will have to dig deeper.

Adam-it:
> Hi Adam, you're right, there seems to be a sort of hickup. It seems that the setup right now "still lives in the past". I did some testing and this are the results:
> 
> * run validate: no code tour selectable
> * run validate: validate code tour selectable
> * run upgrade: validate code tour selectable
> * run validate: validate & upgrade code tour selectable.
> 
> So it seems that only the tour from the previous run is selectable. Timeout doesn't seem to solve the problem right away. Will have to dig deeper.

I wonder if we could open a issue in the code tour extension repo to ask for some advice 🤔?

nicodecleyre:
> > Hi Adam, you're right, there seems to be a sort of hickup. It seems that the setup right now "still lives in the past". I did some testing and this are the results:
> > 
> > * run validate: no code tour selectable
> > * run validate: validate code tour selectable
> > * run upgrade: validate code tour selectable
> > * run validate: validate & upgrade code tour selectable.
> > 
> > So it seems that only the tour from the previous run is selectable. Timeout doesn't seem to solve the problem right away. Will have to dig deeper.
> 
> I wonder if we could open a issue in the code tour extension repo to ask for some advice 🤔?

It's not a tour related issue it seems. I've tested with 100, 200 & 300 ms which didn't do the trick. But with 500 ms latency, it seems to work (tested aaaaaaa lot 😜 ) and 500 is still acceptable for end users i guess. Will update the PR

Adam-it (approved):
Awesome work 👏👏👏👏
I added a minor fixup in the code flow
Tested it locally and works really good 👍 You Rock 🤩

Issue #331, opened 2024-10-21
💡 [Feature]: Improve the general spfx toolkit chat responses to direct to correct chat command


Adam-it:
### 🎯 Aim of the feature

Based on feed I got the general spfx toolkit chat response is not giving directions to the correct chat command: `/new`, `/setup`, `/code`. We should fine-tune it to understand the user intent and if for example the prompt is regarding coding help or guidance the chat participant should suggest to use `/code` command instead of providing answer. It should be clear that the general chat is to provide general feed and information about SPFx and specific scenarios like: setup workspace, scaffodling new project, coding should be handled by mentioning a specific chat action that is grounded with context for this use case.

I should double check if for general chat we include prompt history 
```
// define a chat handler
const handler: vscode.ChatRequestHandler = async (request: vscode.ChatRequest, context: vscode.ChatContext, stream: vscode.ChatResponseStream, token: vscode.CancellationToken) => {

    // initialize the prompt and model
    let prompt = BASE_PROMPT;

    const [model] = await vscode.lm.selectChatModels(MODEL_SELECTOR);

    // make sure the model is available
    if (model) {
        // initialize the messages array with the prompt
        const messages = [
            vscode.LanguageModelChatMessage.User(prompt),
        ];

        // get all the previous participant messages
        const previousMessages = context.history.filter(
            (h) => h instanceof vscode.ChatResponseTurn
        );

        // add the previous messages to the messages array
        previousMessages.forEach((m) => {
            let fullMessage = '';
            m.response.forEach((r) => {
                const mdPart = r as vscode.ChatResponseMarkdownPart;
                fullMessage += mdPart.value.value;
            });
            messages.push(vscode.LanguageModelChatMessage.Assistant(fullMessage));
        });

        // add in the user's message
        messages.push(vscode.LanguageModelChatMessage.User(request.prompt));

        // send the request
        const chatResponse = await model.sendRequest(messages, {}, token);

        // stream the response
        for await (const fragment of chatResponse.text) {
            stream.markdown(fragment);
        }
    }

    return;
};
```

I should double check how we may better understand user context 
https://code.visualstudio.com/updates/v1_94#_drag-and-drop-files-to-add-chat-context

https://code.visualstudio.com/updates/v1_94#_improved-temporal-context-experimental

I should research auto chat participant detection and how SPFx Toolkit may be prepared for it
https://code.visualstudio.com/updates/v1_93#_automatic-chat-participant-detection-in-chat-view-experimental

[...]

PR #341, opened 2024-11-12
Adds manage SharePoint Online resources GH Copilot Chat command. Closes #255, #331

### LINKED ISSUE #255: 💡 [Feature]: new `/manage` for the chat participant
### 🎯 Aim of the feature

The aim is to add a new `/manage` command to the chat participant.
The idea is that it will allow to correlate user prompt with a related CLI command and if possible (has the required inputs for the options) will execute the command to retrieve data in the chat response. It should only base on `get` and `list` CLI command and for start only aim app and SPFx and SPO management

_No response_

follow up #239

Adam-it:
## 🎯 Aim

The aim of this PR is to:

- improve general chat prompt to guide the user to the use the correct chat command for the user intent
- add a new `/manage` command that will allow to retrieve data from SharePoint

## 📷 Result

https://github.com/user-attachments/assets/7a1f1994-95c9-45ca-9015-94fb1a58c715

https://github.com/user-attachments/assets/24a722a7-3770-4252-be14-2a63f818c7d4

## ✅ What was done

- [x] improved general chat prompt
- [x] added manage chat command 
- [x] added script to parse CLI commands to create grounding data for list and get commands for LLM
- [x] added separate LLM request to explain the CLI response 
- [x] improved transfer to `/code` chat command to include previous history context

## 🔗 Related issue

Closes #255, #331


Adam-it:
I should update the readme as well

nicodecleyre:
Hi Adam,

Really, really cool feature! 👏 🚀 

Here's my 2 cent:
- When asking what app catalogs i have (good example btw 😉), i see in the underlying output that he does the `m365 spo tenant appcatalogurl get` command, and gets a result from it (but eventually doesn't do anything with it), and he also executes the `m365 spo site appcatalog list` command and gets a result. Eventually he says the right answer, but maybe the user meant the tenant app catalog? So my point here is, maybe we can build something that when it's not really clear, that the chat asks what the user means. Like:
"Do you want to list the tenant app catalog or all the app catalog sites in your tenant" before proceeding.
- The remove and new feature would really be an added value. But we have to think how we can handle that best. You want the chat to ask all the needed parameters one by one? Or maybe we can build something like an adaptive card where the user has to input all the information when adding something new?

Adam-it:
> * When asking what app catalogs i have (good example btw 😉), i see in the underlying output that he does the `m365 spo tenant appcatalogurl get` command, and gets a result from it (but eventually doesn't do anything with it), and he also executes the `m365 spo site appcatalog list` command and gets a result. Eventually he says the right answer, but maybe the user meant the tenant app catalog? So my point here is, maybe we can build something that when it's not really clear, that the chat asks what the user means. Like:
>   "Do you want to list the tenant app catalog or all the app catalog sites in your tenant" before proceeding.

Good feed 👍. I would need to think how to go about it.

> * The remove and new feature would really be an added value. But we have to think how we can handle that best. You want the chat to ask all the needed parameters one by one? Or maybe we can build something like an adaptive card where the user has to input all the information when adding something new?

This is also a good comment 👍. Currently the GitHub Copilot Chat does not support anything else then action buttons which can trigger any action, also showing some new UI in a new window in editor that present some webview. So this is the only way we could go about it as of now. GH Copilot does not support anything like Adaptive Cards like MS Copilot. 
As of now I was thinking that GH Copilot would ask for missing options that are required to do an action. This is the kind of approach already done in `/new` chat command and seems to be kinda working but the `yo` scaffolding command is quite simple 🤔

Issue #349, opened 2024-11-19
💡 [Feature]: removing an app from app catalog should have a confirmation prompt


Adam-it:
### 🎯 Aim of the feature

The aim is to add and confirmation prompt, a yes/no question, to confirm we are sure we want to remove an app from an app catalog. Currently it will not prompt and just remove the app.

1. We should add this additional notify in this method https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/src/services/actions/CliActions.ts#L182
2. We  should use `showQuickPick` notify like this
```
const shouldRemove = await window.showQuickPick(['Yes', 'No'], {
      title: 'Are you sure you want to remove the app from app catalog?',
      ignoreFocusOut: true,
      canPickMany: false
    });

    const shouldRemoveAnswer = shouldRemove === 'Yes';
```

### 📷 Images (if possible) with expected result

![image](https://github.com/user-attachments/assets/c4b48743-da74-44c3-8e91-9577c3e33ebc)

### 🤔 Additional remarks or comments

parent issue #324

Adam-it:
> Can I work on it?

All yours 🤩👏

Adam-it:
PR merged. This will be live as a new pre-release soon.
Thank you once again for helping out in making the best SPFx tooling in the world 🚀

PR #356, opened 2024-11-20
Prompts for confirmation before removing an app from app catalog. 

Saurabh7019:
## 🎯 Aim

> Prompts for a confirmation before removing an app from the app catalog to prevent accidental removals and improve user experience.

## 📷 Result

> image

## ✅ What was done

- [X] Updated `removeAppCatalogApp` function to check the user's response and proceed with removal only if confirmed.

## 🔗 Related issue

> Closes: #349

### DIFF
--- a/src/services/actions/CliActions.ts
+++ b/src/services/actions/CliActions.ts
@@ -192,6 +192,18 @@ export class CliActions {
 
       const [appID, appTitle, appCatalogUrl] = actionNode.command.arguments;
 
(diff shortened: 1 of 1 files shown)


Adam-it:
Wow that was fast 😮😮😮😮
Thanks for yet another contribution 👏👏👏 
I will try to review it ASAP

Adam-it (approved):
Perfect 🤩👏👏👏👏

Issue #351, opened 2024-11-19
💡 [Feature]: Allow to upgrade an app on a specified site collection using a action in the app catalog app list view


Adam-it:
### 🎯 Aim of the feature

Currently we have a few actions available for each app in the app catalog list. 
The idea/aim is to add a new action which under the hood will use CLI for Microsoft 365 command:
- Upgrade - this will upgrade an app from the specified app catalog in the site. Under the hood it should use [m365 spo app upgrade](https://pnp.github.io/cli-microsoft365/cmd/spo/app/app-upgrade). When clicking on this action we should add prompt for `siteUrl` which is needed so that we will know to which site we should upgrade the selected app

Tips how to go about it, which should be done similar like already available actions
- we should define it as a new app action like this https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/src/constants/Commands.ts#L56-L62
- we should add a new option for it visible in the menu under the app like this https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/src/panels/CommandPanel.ts#L210-L216
- we should handle this new action like this https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/src/services/actions/CliActions.ts#L54-L56 and add logic for the new method like this https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/src/services/actions/CliActions.ts#L182
- The yes/no prompt may be done similar to this https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/src/services/actions/CliActions.ts#L539C5-L545C69
- we should also update the readme 

### 📷 Images (if possible) with expected result

current state of actions for each app in the app catalog list
![image](https://github.com/user-attachments/assets/d7638d10-c52a-48b8-b36c-f50d2376e4e3)

### 🤔 Additional remarks or comments

parent issue #324

Saurabh7019:
Can I take it?

Adam-it:
> Can I take it?

All yours

Adam-it:
PR merged. This will be live as a new pre-release soon.
Thank you once again for helping out in making the best SPFx tooling in the world 🚀

PR #357, opened 2024-11-21
Upgrades an app to the latest version available in the app catalog

Saurabh7019:
## 🎯 Aim

> Upgrades an app to the latest version available in the app catalog for the specified site

## 📷 Result

> image

## ✅ What was done

- [X] New action to upgrade an app in the specified site.
- [X]  Prompt for the site URL only when upgrading tenant app catalog apps. For site collection app catalog apps, it uses the site collection app catalog URL and scope.

## 🔗 Related issue

> Closes: #351

### DIFF
--- a/src/panels/CommandPanel.ts
+++ b/src/panels/CommandPanel.ts
@@ -211,7 +211,8 @@ export class CommandPanel {
                     new ActionTreeItem('Retract', '', undefined, undefined, Commands.retractAppCatalogApp, [app.ID, app.Title, undefined, app.Deployed], ContextKeys.retractApp),
                     new ActionTreeItem('Remove', '', undefined, undefined, Commands.removeAppCatalogApp, [app.ID, app.Title], ContextKeys.removeApp),
                     new ActionTreeItem('Enable', '', undefined, undefined, Commands.enableAppCatalogApp, [app.Title, tenantAppCatalogUrl, app.Enabled], ContextKeys.enableApp),
-                    new ActionTreeItem('Disable', '', undefined, undefined, Commands.disableAppCatalogApp, [app.Title, tenantAppCatalogUrl, app.Enabled], ContextKeys.disableApp)
+                    new ActionTreeItem('Disable', '', undefined, undefined, Commands.disableAppCatalogApp, [app.Title, tenantAppCatalogUrl, app.Enabled], ContextKeys.disableApp),
+                    new ActionTreeItem('Upgrade', '', undefined, undefined, Commands.upgradeAppCatalogApp, [app.ID, app.Title], ContextKeys.upgradeApp)
                   ]
                 )
               );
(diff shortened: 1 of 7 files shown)


Adam-it:
Wow... Awesome!. You Rock 🤩

src/services/actions/CliActions.ts
```diff
@@ -224,6 +227,53 @@ export class CliActions {
...
   }
 
+  /**
+  * Upgrades an app to a newer version available in the app catalog.
+  *
+  * @param node The tree item representing the app to be upgraded.
+  */
+  public static async upgradeAppCatalogApp(node: ActionTreeItem) {
+    try {
+      const actionNode = node.children?.find(child => child.contextValue === ContextKeys.removeApp);
```
Adam-it: shouldn't this be `upgradeApp`

```suggestion
      const actionNode = node.children?.find(child => child.contextValue === ContextKeys.upgradeApp);
```

src/panels/CommandPanel.ts
```diff
@@ -211,7 +211,8 @@ export class CommandPanel {
                     new ActionTreeItem('Retract', '', undefined, undefined, Commands.retractAppCatalogApp, [app.ID, app.Title, undefined, app.Deployed], ContextKeys.retractApp),
                     new ActionTreeItem('Remove', '', undefined, undefined, Commands.removeAppCatalogApp, [app.ID, app.Title], ContextKeys.removeApp),
                     new ActionTreeItem('Enable', '', undefined, undefined, Commands.enableAppCatalogApp, [app.Title, tenantAppCatalogUrl, app.Enabled], ContextKeys.enableApp),
-                    new ActionTreeItem('Disable', '', undefined, undefined, Commands.disableAppCatalogApp, [app.Title, tenantAppCatalogUrl, app.Enabled], ContextKeys.disableApp)
+                    new ActionTreeItem('Disable', '', undefined, undefined, Commands.disableAppCatalogApp, [app.Title, tenantAppCatalogUrl, app.Enabled], ContextKeys.disableApp),
+                    new ActionTreeItem('Upgrade', '', undefined, undefined, Commands.upgradeAppCatalogApp, [app.ID, app.Title], ContextKeys.upgradeApp)
```
Adam-it, line 215: I just had an awesome idea for an improvement that we could also pass here the tenant Url so user may just provide the rest of the url or only the relative part of the url. So for exmaple in my case the site is 
`https://tenanttocheck.sharepoint.com/sites/PnPDemo2` and I need to provide the full url. But SPFx Toolkit is already aware of my tenant Url so `https://tenanttocheck.sharepoint.com/`. It would be awesome if the user could only provide `sites/PnPDemo2`. What do you think?
I say let's not do that now but create a follow up issue for an improvement in this feature

Adam-it (approved):
Works really good 👏👏👏👏
The code looks perfect 🤩 I only added a small fixup

PR #358, opened 2024-11-22
Improves re-login experience. Closes #342, #374

### LINKED ISSUE #342: 💡 [Feature]: Improve re-login experience
### 🎯 Aim of the feature

I noticed that SPFx toolkit has a glitch that even though the status command returns that I am logged in, the first command to get the app catalogs from my tenant requires me to sign in which is only visible in the output window. This leads to a state were I am not logged in, I don't see anything in the account and app catalog view, but I also don't see the sign-in button. We should fix it up.

From what i checked the token expired and I need to resign to refresh it. This is fine but:
 - first is the above described problem, I don't see the sign it or out button, I need to 'manually' sign out using the command palette in VS Code and then I am able to relogin
 - when signing out we clear out the client Id and tenant Id. Usually I don't remember those so I just create an another entra app reg. I must have like 40 SPFx Toolkit apps by now on my tenant 😅. What we could do is 'cache' the client and tenant Id so that when next time we relogin and we pick to sign in using the existing client Id and tenant Id we have those fields prefilled with the previous values (when we previously sign in). VS Code has a mechanism for that to preserve state between multiple sessions of VS Code and we could use that. 

_No response_

_No response_

### Prototype 

1. in global handle of any CLI command execution we need to add logic that when a login message will get present we should present a notify to the user to relogin to your tenant and sign out user totally and show like initial state
2. we should persist the ClientID and TenantID in VS Code global state and overwrite it every time we login.
https://stackoverflow.com/questions/51821924/how-to-persist-information-for-a-vscode-extension
based on what we have in the global state we should prefill the clientId and tenantId in the login notifys

Adam-it:
## 🎯 Aim

The aim improve and simplify login behavior for SPFx Toolkit and try to fixup a strange issue with login output when cli command is executed.

## ✅ What was done

- [x] updated login method from device code to browser to simplified the process
- [x] added saving clientId and tenantId to global state

## 🔗 Related issue

Closes #342, #374

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -23,7 +23,7 @@ export async function activate(context: vscode.ExtensionContext) {
 
 	TerminalCommandExecuter.register();
 
-	AuthProvider.register();
+	AuthProvider.register(context);
 
 	Dependencies.registerCommands();
 	Scaffolder.registerCommands();
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -28,7 +28,7 @@ export async function activate(context: vscode.ExtensionContext) {
 	Dependencies.registerCommands();
 	Scaffolder.registerCommands();
 	CliActions.registerCommands();
-	EntraAppRegistration.registerCommands();
+	EntraAppRegistration.registerCommands(context);
 
 	CommandPanel.register();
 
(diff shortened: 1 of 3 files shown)


Adam-it (commented):
- [ ] I should update the docs.
- [x] I should add notify when token expires with relogin message. Now it is present only in extension output

Adam-it:
> * [ ]  I should update the docs.
> * [x]  I should add notify when token expires with relogin message. Now it is present only in extension output

I will update the readme when doring a minor release as it requires more restructuring and work

PR #387, opened 2025-01-03
🚀 Add command to increase project version and implement version increment logic, Closes #94

### LINKED ISSUE #94: 💡 [Feature]: Extra tasks to update Patch, Minor and Major version of SPFx project
## 🎯 Aim of the feature

Add extra tasks so that you can update the patch, minor or major version of an SPFx project (package-solution.json) in one click. 
Extra nice would be if it would also update the package.json version as well.

## 📷 Images (if possible) with expected result

image

nicodecleyre:
## 🎯 Aim

This pull request increases the version of the package.json & package-solution.json file with either major, minor or patch 

## 📷 Result

https://github.com/user-attachments/assets/cd47dd06-699f-4fae-a313-30b54c39e0ed

## ✅ What was done

- [X] added button to increase version
- [X] created prompt that asks for 'major', 'minor' or 'patch' when clicking on the button
- [X] increasing the number with function in separate file
- [X] Change the increased number in package.json
- [X] Change the increased number in package-solution.json (with an extra '0' at the end for the revision so the format is X.X.X.0)

## 🔗 Related issue

Closes: #94


src/utils/increaseVersion.ts
```diff
@@ -0,0 +1,39 @@
...
+
+
+export async function increaseVersion(versionType: 'major' | 'minor' | 'patch') {
+  const wsFolder = workspace.workspaceFolders?.[0];
+  if (!wsFolder) {
+    throw new Error('Workspace folder not found');
+  }
+
+  const packageJsonPath = join(wsFolder.uri.fsPath, 'package.json');
+  const packageSolutionPath = join(wsFolder.uri.fsPath, 'config/package-solution.json');
```
Adam-it, line 13: this will not work when the SPFx project was created by TeamsToolkit. We should also support that case. 
In case TT created a project that uses SPFx project then the SPFx part is in a `src` subcatalog. 
I took a screenshot of the result when we use SPFx Toolkit in a project created by TT.
image

We already have some parts in SPFx Toolkit that add this support and check. Like here
https://github.com/pnp/vscode-viva/blob/beaf740d93408a4d8803272282ce2814bdd7cc2a/src/services/check/DebuggerCheck.ts#L37-L41

src/panels/CommandPanel.ts
```diff
@@ -302,6 +306,7 @@ export class CommandPanel {
     actionCommands.push(new ActionTreeItem('Upgrade project SPFx version', '', { name: 'arrow-up', custom: false }, undefined, Commands.upgradeProject));
     actionCommands.push(new ActionTreeItem('Validate project correctness', '', { name: 'check-all', custom: false }, undefined, Commands.validateProject));
     actionCommands.push(new ActionTreeItem('Rename project', '', { name: 'whole-word', custom: false }, undefined, Commands.renameProject));
+    actionCommands.push(new ActionTreeItem('Increase project version', '', { name: 'arrow-up', custom: false }, undefined, Commands.increaseVersion));
```
Adam-it, line 309: lets pick a different icon. This one is already used in upgrade action which may be just confusing.
we may use any icon from this 
https://microsoft.github.io/vscode-codicons/dist/codicon.html
what do you think about `fold-up` icon 🤔

Adam-it (changes requested):
@nicodecleyre awesome work so far. I added some small fixup to your branch. I left also two comments that we could recheck before we merge.

src/utils/increaseVersion.ts
```diff
@@ -0,0 +1,44 @@
+import { readFileSync, writeFileSync } from 'fs';
+import { join } from 'path';
+import { workspace } from 'vscode';
+
+export async function increaseVersion(versionType: 'major' | 'minor' | 'patch') {
```
Adam-it, line 6:
```suggestion


export async function increaseVersion(versionType: 'major' | 'minor' | 'patch') {
```

Adam-it (commented):
I exposed this as a new VS code command so we may also start it using the VS Code commands when using F1 key
and I performed a bit of clean-up
LGTM

Adam-it:
@nicodecleyre I did the review and I have one last thing I noticed I wanted to recheck with you. 
When running this command in a SPFx project as port of TeamToolkit project we actually have 3 places with version 🤦‍♂️
We have the 
- `package-solution.json` which is part of the spfx project inside the nested `src` folder
- `package.json` which is also part of spfx project in the nested `src` folder
- `package.json` which is also part of teams toolkit project

image

From what I checked the `package.json` form the spfx part does not get updated in this case. I think this might be wrong and I think we should do one of the following:
- either in this case we update and align all three of those palces
- or (and I am thinking if this will be better) we only update the version in the `package-solution.json` and `package.json` which are part of the SPFx project and we do not change anything on the Teams Toolkit project.

Which option would you pick? What do you think?

Adam-it:
@nicodecleyre just checking, did you had some time to check my last comment? Any feed on that? 
I would like to sort it out before we merge this PR

nicodecleyre:
> @nicodecleyre I did the review and I have one last thing I noticed I wanted to recheck with you. 
> When running this command in a SPFx project as port of TeamToolkit project we actually have 3 places with version 🤦‍♂️
> We have the 
> - `package-solution.json` which is part of the spfx project inside the nested `src` folder
> - `package.json` which is also part of spfx project in the nested `src` folder
> - `package.json` which is also part of teams toolkit project
> 
> image
> 
> From what I checked the `package.json` form the spfx part does not get updated in this case. I think this might be wrong and I think we should do one of the following:
> - either in this case we update and align all three of those palces
> - or (and I am thinking if this will be better) we only update the version in the `package-solution.json` and `package.json` which are part of the SPFx project and we do not change anything on the Teams Toolkit project.
> 
> Which option would you pick? What do you think?
> 

I think we should update it in all 3 places and build in a catch if json is not present

Adam-it:
> I think we should update it in all 3 places and build in a catch if json is not present

good idea. Let's proceed like this and merge. 👍 
Would you be so kind and make the required changes?

Adam-it:
@nicodecleyre we just did a minor release SPFx Toolkit. Due to that the `dev` branch was overwritten and aligned with `main` branch. 
Please align your `dev` and then rebase your feature branch in order to fixup the git history of this PR. Thanks 👍

nicodecleyre:
> @nicodecleyre we just did a minor release SPFx Toolkit. Due to that the `dev` branch was overwritten and aligned with `main` branch. Please align your `dev` and then rebase your feature branch in order to fixup the git history of this PR. Thanks 👍

Hi @Adam-it, sorry for the long wait, was pretty busy last month. I rebased dev, can you check if it's ok? Thank youu 🙏

Adam-it:
> Hi @Adam-it, sorry for the long wait, was pretty busy last month. I rebased dev, can you check if it's ok? Thank youu 🙏

Totally understandable. You Rock man🤩! don't you ever forget about that 👍
I will try to take a look at it today.

Adam-it (approved):

Adam-it (approved):
Awesome work 👏👏👏👏

PR #389, opened 2025-01-12
Adds support for MS Teams in GitHub Copilot `/manage` chat command

Adam-it:
## 🎯 Aim

The aim is to extend the list of supported commands/skills with teams area

The biggest problem I see with this is that we need to extend the SPFx Toolkit app reg with additional scopes which are needed only for this functionality. I am not sure if this is a bit of over kill for this tool

## 📷 Result

image

image

image

## ✅ What was done

- [X] Added more commands
- [X] Small update in the chat handler
- [ ] Extend docs with additional scopes

## 🔗 Related

https://github.com/Adam-it/cli-microsoft365/commit/1fef118f9ca099023434bd98cce5fd0a95d8d3d2

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,12 +1,12 @@
 {
 	"name": "viva-connections-toolkit",
-	"version": "4.3.0",
+	"version": "4.3.1",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
 		"": {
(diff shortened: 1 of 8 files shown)


Adam-it:
lets drop this for now and try to improve spo area and focus on this issue #394
I will leave this change on my fork so we may always get back to this issue if needed

PR #390, opened 2025-01-15
Install, Uninstall apps to a specified site. Closes #350

### LINKED ISSUE #350: 💡 [Feature]: Allow to install/uninstall an app on a specified site collection using a action in the app catalog app list view
### 🎯 Aim of the feature

Currently we have a few actions available for each app in the app catalog list. 
The idea/aim is to add two new actions which under the hood will use CLI for Microsoft 365 commands:
- Install - this will install an app from the specified app catalog in the site. Under the hood it should use [m365 spo app install](https://pnp.github.io/cli-microsoft365/cmd/spo/app/app-install). When clicking on this action we should add prompt for `siteUrl` which is needed so that we will know to which site we should install the selected app
- Uninstall - this will uninstall an app from the specified site. Under the hood it should use [m365 spo app uninstall](https://pnp.github.io/cli-microsoft365/cmd/spo/app/app-uninstall). When clicking on this action we should add prompt for `siteUrl` which is needed so that we will know to from which site we should uninstall the selected app. We should also prompt for confirmation with a yes/no- notify if the user is sure to uninstall the app.

Tips how to go about it, which should be done similar like already available actions
- we should define it as a new app action like this https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/src/constants/Commands.ts#L56-L62
- we should add a new option for it visible in the menu under the app like this https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/src/panels/CommandPanel.ts#L210-L216
- we should handle this new action like this https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/src/services/actions/CliActions.ts#L54-L56 and add logic for the new method like this https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/src/services/actions/CliActions.ts#L182
- The yes/no prompt may be done similar to this https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/src/services/actions/CliActions.ts#L539C5-L545C69
- we should also update the readme 

current state of actions for each app in the app catalog list 
image

parent issue #324

Saurabh7019:
## 🎯 Aim

Allows to install/uninstall an app on a specified site collection using a action in the app catalog app list view

## 📷 Result

image

## ✅ What was done

- [X] Install/Uninstall actions added to both tenant and site collection app catalog apps
- [X] Prompts for relative site URL when actions triggerred in tenant app catalog apps. 
- [X] Requires confirmation before uninstalling app.

## 🔗 Related issue

Closes: #350


Adam-it (approved):
Perfect 🤩
You Rock 👍
Checked locally ✅

Issue #217, opened 2024-04-17
💡 [Feature]: Add possibility to add custom additional step


Adam-it:
### 🎯 Aim of the feature

We should add some possibilities, preferably extending it as a setting for the extension, to allow devs to define their own additional setup options. 
We could expose a setting in which users could manage a json array that will have a `label` and `command`. `label` will hold the string that will be present in the UI in the scaffolding form and `command` will be the npm command that should be executed after this option was selected.

example idea for a this kind of json object:
```json
[
  {
    "label": "PnP JS SharePoint",
    "command": "npm install @pnp/sp"
  }
]
```

We should NOT allow to disable the pre-defined options

### 🤔 Additional remarks or comments

doing this change we should keep an eye out for this issue #216
if it is already done we should also add a option like `visiblity` that will allow to list out in as string array components that should have this option visible

ervingayle:
Hi @Adam-it. How did you envision the array being stored?

ervingayle:
Hi @Adam-it . Thank you for the example. This helps me. I did not know of this option. I know that you mentioned not allowing for removal of the default ones but this json array would only specify the additional packages that someone would want installed. Yes, I believe that this addition would require good docs with examples to effectively help everyone.

ervingayle:
Yes, please assign this to me.

Adam-it:
> Yes, please assign this to me.

All yours 👍

ervingayle:
Hi @Adam-it. I have made progress with this item but I do have some issues with the implementation. Currently I tried with selecting React and React Dom, PnP and then I added a new "Additional Step" called Install custom steps. This reads the setting value and then I loop through the json. This works. However, what I am seeing is that the React and React Dom step is being bundled with the install dependencies initial powershell task where npm i is executed. Every additional custom step is created in a new terminal window and installed simultaneously. In some cases the package is installed and no errors are shown, folder in node_modules but the package entry is not in package.json. I have tested this about 15 times so far. Do you know what would cause this?

Also, I intended to show what would be installed in the scaffolding form but passing the setting value to the view didn't work as I expected so for now I have reverted that code.
I have not opened a pull request as yet. Code [here](https://github.com/ervingayle/vscode-viva/)

Adam-it:
@ervingayle thanks for reaching out. 
[...]

Issue #325, opened 2024-10-15
💡 [Feature]: Remove or Refactor the show/hide tenant wide extension list


Adam-it:
### 🎯 Aim of the feature

Since now the list of extensions only loads when the node is expanded to ensure that the additional API calls don't affect the extension load time there is little sense in having the setting that allows to show/hide the extension list. We could consider the following:
- removing this setting. This may seem a breaking change and maybe we should consider doing it with next major v5 release which should also bring multiple connections. This release is a bit blocked due to ESM support (different issue/story 😛)
- refactoring this setting to something more global. It could now allow to either show(load)/hide apps from all app catalogs. So when set to true all app catalogs will be expandable and will show the app list. When set to true we would only show app catalogs without the possibility to show apps

### 📷 Images (if possible) with expected result

The issue relates to the following setting:
![image](https://github.com/user-attachments/assets/cc77a0b6-902f-440d-b139-36769c5760f1)

### 🤔 Additional remarks or comments

Related issue #215

Tagging @Saurabh7019 as initial author of this idea 👍. You Rock 🤩

### Prototype

1. we should rename this setting and correct its description
https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/package.json#L187-L191
it should be something like `Show apps in app catalogs` and description like `When set it will load all apps in every app catalog on your tenant`
2. we should remove this https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/src/panels/CommandPanel.ts#L176 and just always load the tenant wide extensions
3. we should use the new setting here https://github.com/pnp/vscode-viva/blob/59190115e0c676068a3f1a3698e31f3500aee3f6/src/panels/CommandPanel.ts#L200 and it similar place in site app catalogs so that when this setting is set to false we should not make it expandable and not add the expand handle
4. this setting should be `true` by default

nicodecleyre:
Hi, @Adam-it, just wondering. You define in the last bullet point 2 scenarios 'when set to true'. I was wondering if it shouldn't be:
- When set to true, all app catalogs will be expandable and display the app lists.
- When set to false, only the catalogs themselves will be visible, without the option to expand and view individual apps.

Adam-it:
> Hi, @Adam-it, just wondering. You define in the last bullet point 2 scenarios 'when set to true'. I was wondering if it shouldn't be:
[...]

PR #404, opened 2025-01-30
Add possibility to add custom additional step

ervingayle:
## 🎯 Aim

This PR implements functionality that allows for custom npm packages to be installed as part of the scaffolding project setup.

## 📷 Result

> include images or a movie with a result of your change [Remove this line]
vscode-viva-customstepsetting-1
vscode-viva-customstepsetting-2
vscode-viva-customstepsetting

## ✅ What was done

- [X] Added new extension setting for custom steps
- [X] Added custom toggle to activate the
- [X] Added logic to iterate over extension setting value json array and object(s)
- [X] Added notice on how to access the setting

## 🔗 Related issue

Closes: #217


Adam-it:
I love the ID of this PR 😄 404
I would love to review it but I just simply cannot find it 😂😁

ervingayle:
It would be my PR that gets this 404 ID.

src/extension.ts
```diff
@@ -85,6 +86,20 @@ export async function activate(context: vscode.ExtensionContext) {
...
 					await TerminalCommandExecuter.runCommand(nodeVersionCommand, terminalTitle, terminalIcon);
 				}
+
+				//Look for the custom steps string in the project file and run the command if found
+				if (fileContents.indexOf(ProjectFileContent.installCustomSteps) > -1) {
+					const value = getExtensionSettings<string>('projectCustomSteps', '');
+					if (value) {
+						const jsonArray = JSON.parse(value);
+						// eslint-disable-next-line no-console
+						console.log(jsonArray.length);
```
Adam-it, line 96: we should avoid console logs in code at all time. If we want to add additional logging we should use `logger` for that

src/extension.ts
```diff
@@ -85,6 +86,20 @@ export async function activate(context: vscode.ExtensionContext) {
...
+						for (let i = 0; i < jsonArray.length; i++) {
+							const item = jsonArray[i];
+							await TerminalCommandExecuter.runCommand(item.command, terminalTitle, terminalIcon);
```
Adam-it, line 99: let's consolidate this to a single line. If `item` is used in a single place the there is little benefit to extract it to a separate variable

```suggestion
							await TerminalCommandExecuter.runCommand(jsonArray[i].command, terminalTitle, terminalIcon);
```

src/panels/CommandPanel.ts
```diff
@@ -171,31 +171,35 @@ export class CommandPanel {
...
-            if (tenantWideExtensions && tenantWideExtensions.length > 0) {
-              tenantWideExtensions.forEach((extension) => {
-                tenantWideExtensionsList.push(
-                  new ActionTreeItem(extension.Title, '', { name: 'spo-app', custom: true }, TreeItemCollapsibleState.None, 'vscode.open', Uri.parse(extension.Url), 'sp-app-catalog-tenant-wide-extensions-url')
-                );
-              });
-            } else {
-              tenantWideExtensionsList.push(new ActionTreeItem('No extension found', ''));
-            }
+      const tenantWideExtensionsNode = new ActionTreeItem('Tenant-wide Extensions', '', { name: 'spo-app-list', custom: true }, TreeItemCollapsibleState.Collapsed, undefined, undefined, 'sp-app-catalog-tenant-wide-extensions', undefined,
```
Adam-it, line 174: wooops! Seems like those changes should not be part of this PR as they are not needed for this issue. 
We should clean this up before we proceed.

Adam-it (changes requested):
@ervingayle Awesome work so far 👏👏👏 
I had to stop my review as I noticed that you pushed commits to this branch that are part of a different issue you are assigned to. I think the problem is you were working on your `dev` branch instead of separate branch based on `dev` that was created only for this feature. 
We should clean this up before we proceed. Best approach would be to create new branches and cherry-pick to them only the commits that are required for issue #217 and close this PR without merging and reopen an another PR from a different branch then this one. 
I added more context in the issue #325 which might be helpful how to get around that

package.json
```diff
@@ -232,6 +232,13 @@
 						"code tour"
 					],
 					"description": "Choose the output mode for the project validation. You can choose between `both`, `markdown`, or `code tour`."
+				},
+				"spfx-toolkit.projectCustomSteps": {
+					"title": "Additional Custom Steps",
+					"type": "string",
+					"default": "[\n  {\n    \"label\": \"PnP JS SharePoint\",\n    \"command\": \"npm install @pnp/sp\"\n  },\n  {\n    \"label\": \"PnP React Reusable Controls\",\n    \"command\": \"npm install ....\"\n  }\n]",
```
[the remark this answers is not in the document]
ervingayle: I thought that a sample should be provided by default even if it is the sample that was posted in the initial issue. However, I will follow your suggestion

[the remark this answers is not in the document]
Adam-it: maybe we could add some link to wiki with more guidance in the label. This is the settings value which I guess will fail if we have `....` in it. I guess it wont be valid right

Adam-it:
@ervingayle I will close this PR and let's open it up again once we managed to sort out the git history problems

PR #405, opened 2025-02-04
Release v4.4.0

Adam-it:
## 🎯 Aim

The aim is to create a new minor release with the latest enhancements. 

## 📷 Result

image

image

image

image

## ✅ What was done

- [X] Unfied the gulp bundle local and production to a single task with an additional prompt
- [X] Unfied the gulp package-solution local and production to a single task with an additional prompt
- [X] Fixed warning message regarding installed Node.js version in validate dependency action
- [X] Added conditional logic to scaffolding form additional dependencies to show them based on project type
- [X] Added install react additional dependency to the scaffolding form visible for Application customizer
- [X] Added Install, and Uninstall apps to a specified site action to app view
- [X] Enhanced progress notifications with output window links for better user guidance to every CLI action
- [X] Added action which allows to set form customizer 
- [X] Enhanced error handling in GitHub Copilot Chat participant
- [X] Updated welcome view
- [X] Updated VS Code typings and engine to latest


Adam-it (commented):
✅

Issue #279, opened 2024-07-27
💡 [Feature]: for the love of god let's start writing tests


Adam-it:
### 🎯 Aim of the feature

It's kinda embarrassing but yes we do not have unit tests as of now.
What initially started as a small VS Code extension oriented around  ACEs and Viva Connections turned to this massive extension that extensively enhances VS Code with lot of SPFx development features, now even extends GitHub Copilot.
As we grow and grow we need to make sure we increase the quality and maintainability of our work. As first step we need to add unit tests.
The aim of this issue is to prototype this and kick of with a small portion of unit tests for a single provider or service.

ervingayle:
Hi @Adam-it. I did some research on this and see that it is possible to do some unit tests with Mocha or with a third party vscode-extension-tester that appears to be more suited to test aspects of an extension. What are your thoughts on approach? I would like to work on this implementation starting with prototype.

Adam-it:
> Hi @Adam-it. I did some research on this and see that it is possible to do some unit tests with Mocha or with a third party vscode-extension-tester that appears to be more suited to test aspects of an extension. What are your thoughts on approach? I would like to work on this implementation starting with prototype.

Thanks @ervingayle for your initiative 👏👍. 
Unfortunately this issue is still not ready to be taken (it is not labeled as 'help wanted') and I am still thinking and prototyping on the best approach to go about this task. 
TBH for the last couple of months I haven't found time fot this yet 🫤, there is always so much to do 😉. 
That being said I am 100% open for feedback and suggestions. Till now I haven't picked an approach yet so if you had anything specific in mind or would like to point me to an example or article or tech you would use I would really appreciate it.
Till now I considered just writing standard unit tests using Mocha as I have some experience in it and this would work perfectly for services like CliActions.
On the other hand many parts of this extension is not ready for testing, like those that extend VS Code UI with web views, and here I wanted to do something like integration testing that would run the extension based on a scenario but top priority is to make it work in GH Actions.

@ervingayle do you want to share your research?

ervingayle:
[...]

PR #414, opened 2025-02-22
Adds testing prototype solution. Related #279

Adam-it:
## 🎯 Aim

The aim is to prototype integration tests for the solution

To debug tests:
1. install extension https://marketplace.visualstudio.com/items?itemName=ms-vscode.extension-test-runner
2. run `npm run package && npm run compile-tests`
3. run `Extension Tests` in debug

To run:
1. run `npm run package && npm run compile-tests`
2. run `npm run test` and check the result in terminal

## 📷 Result

image

## ✅ What was done

- [X] added `test-cli` and `test-electron`
- [X] added a simple test that checks if extension is running

## 🔗 Related issue

Related #279

### DIFF
--- a/src/test/runTests.ts
+++ b/src/test/runTests.ts
@@ -1,6 +1,6 @@
 import * as path from 'path';
-
 import { runTests } from '@vscode/test-electron';
+
 
 async function main() {
 	try {
--- a/src/test/runTests.ts
+++ b/src/test/runTests.ts
@@ -1,0 +1,22 @@
+import * as path from 'path';
+
+import { runTests } from '@vscode/test-electron';
+
+async function main() {
+	try {
+		// The folder containing the Extension Manifest package.json
+		// Passed to `--extensionDevelopmentPath`
+		const extensionDevelopmentPath = path.resolve(__dirname, '../../');
+
+		// The path to the extension test script
+		// Passed to --extensionTestsPath
+		const extensionTestsPath = path.resolve(__dirname, './suite/index');
+
+		// Download VS Code, unzip it and run the integration test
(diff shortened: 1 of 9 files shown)


src/test/runTests.ts
```diff
@@ -0,0 +1,22 @@
+import * as path from 'path';
+
+import { runTests } from '@vscode/test-electron';
```
Adam-it, line 3:
```suggestion
import * as path from 'path';
import { runTests } from '@vscode/test-electron';
```

src/test/runTests.ts
```diff
@@ -0,0 +1,22 @@
...
+
+async function main() {
```
Adam-it, line 5:
```suggestion


async function main() {
```

src/test/suite/index.ts
```diff
@@ -0,0 +1,36 @@
+import * as path from 'path';
+import * as Mocha from 'mocha';
+import * as glob from 'glob';
+
+export function run(): Promise<void> {
```
Adam-it, line 5:
```suggestion
import * as glob from 'glob';


export function run(): Promise<void> {
```

src/test/suite/extension.test.ts
```diff
@@ -0,0 +1,18 @@
+import * as assert from 'assert';
+import * as vscode from 'vscode';
+import { sleep } from '../../utils/sleep';
+
+suite('Extension', () => {
```
Adam-it, line 5:
```suggestion
import { sleep } from '../../utils/sleep';


suite('Extension', () => {
```

PR #438, opened 2025-04-03
support for installing correct version of Node.js. Closes #437

### LINKED ISSUE #437: 💡 [Feature]: Support installing correct version of Node.js
### 🎯 Aim of the feature

Currently what we do is only check if you have Node.js installed and if the Node.js version is correct and if not we just present and error message and that's it. 
We check it when running check dependencies step. 
The idea is to add more support to this part so to guide a user how to install Node.Js or provide some link/action that would install Node.js for the user. The biggest challenge here is that are different ways we install Node.js on Windows (Winget) and Mac/Linux (Curl). So maybe as minimum we could provide a link from which we may download Node.js manually. 
Also we could think of providing some common guidance to use NVM or NVS which SPFx Toolkit supports already.

_No response_

_No response_

Saurabh7019:
## 🎯 Aim

Support installing correct version of Node.js

## 📷 Result

image

## ✅ What was done

- [X] provides a link to download and install Node.js
- [X] provides common guidance to use NVM (based on users' current OS) or NVS

## 🔗 Related issue

Closes: #437


Adam-it (changes requested):
Awesome work 👏

I have only one small comment which I something initially I did not think of. I suggest we now also change the action name from `check dependecies` to something like `validate local setup` or `check workspace`. What do you think?

image

image

image

Adam-it:
> `validate local setup` is better, and it's also in line with what Microsoft learn articles use. But honestly, `check dependencies` wasn't bad either. :)

ok let's go with `validate local setup`

> What about 'Install dependencies'? Should we rename that as well as part of this PR?

I would leave this as what it essentially does it really just runs npm install of the needed tools. So the naming is correct here as well. In case of validate now we not only validate npm package but also Node version and provide guidance how to obtain it

Adam-it:
@Saurabh7019 I know you most probably going to kill me 😅
but due to this https://devblogs.microsoft.com/microsoft365dev/announcing-sharepoint-framework-1-21-with-updates-on-building-enterprise-extensibility-within-microsoft-365/
I did a minor release to add support for SPFx 1.21 to latest stable ASAP and part of this version of SPFx is the support of Node v22 😉
May I super kindly ask you to align (again) 🙏, We need to update the node versions in this PR to 22

Saurabh7019:
No problem at all :) I will update it today.

Saurabh7019:
made a mess while rebasing; it should be fine now.

Adam-it (approved):
@Saurabh7019 Awesome work 👏👏👏👏
Checked locally ✅
The code LGTM 

image


image

PR #459, opened 2025-04-29
Hides app management related context-menu actions from the command pa…

### LINKED ISSUE #423: 💡 [Feature]: Hide the VSCode commands for this extension that will not work
### 🎯 Aim of the feature

The aim is to clear up some of the commands that are present in the command bar for this extension but will not work and their aim was to be present only as options in the context menu for an app.
TBH I am not giving this as `good first issue` as I am not 100% sure if this may be done and this requires a bit of research and experimenting.

So to make it clear. We should update this part to hide the actions we have in the context menu from VSCode command bar

Image

_No response_

Saurabh7019:
## 🎯 Aim

> Hides app management related context-menu actions from the command palette.

## 📷 Result

> before 
image

> After 
image

## ✅ What was done

- [X] Hide all app management related actions from the command palette.

## 🔗 Related issue

Closes: #423

### DIFF
--- a/package.json
+++ b/package.json
@@ -803,6 +803,42 @@
 				{
 					"command": "spfx-toolkit.validateProject",
 					"when": "pnp.project.isSPFxProject"
+				},
+				{
+					"command": "spfx-toolkit.deployAppCatalogApp",
+					"when": "false"
+				},
+				{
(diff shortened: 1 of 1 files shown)


Adam-it (approved):
LGTM 👍

PR #466, opened 2025-05-03
Shorten command category names. Closes #421

### LINKED ISSUE #421: 💡 [Feature]: Make extension name in VS Code commands shorter
### 🎯 Aim of the feature

It is possible to run almost any action or other feature of SPFx Toolkit as a VSCode command. The problem is the extension name is very long and we need to write `SharePoint Framework Toolkit:` before we start specifying which action we want to run. 
So in order to make this easier let's update this name, ONLY for the VSCode commands part, to `SPFx Toolkit:` 

We should update the name of those commands and change the extension name part (so the one marked in blue) to shorter `SPFx Toolkit`

Image

_No response_

Adam-it:
## 🎯 Aim

The aim is to make the VS Code extension commands shorter

## 📷 Result

image

## ✅ What was done

- [X] Updates category name for the VS Code commands

## 🔗 Related issue

Closes: #421

### DIFF
--- a/package.json
+++ b/package.json
@@ -612,177 +612,177 @@
 			{
 				"command": "spfx-toolkit.welcome",
 				"title": "Welcome",
-				"category": "SharePoint Framework Toolkit"
+				"category": "SPFx Toolkit"
 			},
 			{
 				"command": "spfx-toolkit.checkDependencies",
 				"title": "Validate local setup",
-				"category": "SharePoint Framework Toolkit"
+				"category": "SPFx Toolkit"
 			},
 			{
 				"command": "spfx-toolkit.installDependencies",
 				"title": "Install dependencies",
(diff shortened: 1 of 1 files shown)


Saurabh7019 (approved):

Saurabh7019:
image

@Adam-it, all actions are renamed now.. but some commands still show with longer name so just wondering if this could be confusing to some users. Changing this would require changing the view container name, so let's discuss that separately.

Adam-it:
> image
> 
> @Adam-it, all actions are renamed now.. but some commands still show with longer name so just wondering if this could be confusing to some users. Changing this would require changing the view container name, so let's discuss that separately.

Good find. Let's open a follow up issue for that and I would rename those as well 👍

Saurabh7019:
great, I will create a follow-up issue for this.

PR #467, opened 2025-05-03
Update default action for SPFx app to install. Closes #429

### LINKED ISSUE #429: 💡 [Feature]: Update the default action for the SPFx app
### 🎯 Aim of the feature

Currently, the default action for every SPFx app in all app catalogs are `deploy` but this makes little sense. Currently, the apps are almost always deployed for example when we use the deploy action it is already done. 
We should refactor this to `install` action as this is makes more sense. After we add and deploy the app to the app catalog we want to install it to a site. this should be the quick action we should promote

We should update this from deploy to install and also we should use a different icon

Image

_No response_

Adam-it:
## 🎯 Aim

Update default action for SPFx app to install

## 📷 Result

image

## ✅ What was done

- [X] Update default action for SPFx app to install

## 🔗 Related issue

Closes: #429

### DIFF
--- a/package.json
+++ b/package.json
@@ -860,7 +860,7 @@
 					"group": "inline@1"
 				},
 				{
-					"command": "spfx-toolkit.deployAppCatalogApp",
+					"command": "spfx-toolkit.installAppCatalogApp",
 					"when": "view == pnp-view-environment && viewItem == pnp.etv.hasAppCatalogApp",
 					"group": "inline@1"
 				},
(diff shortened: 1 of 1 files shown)


Saurabh7019 (approved):

Saurabh7019:
made a minor change here @adam, the deploy app catalog action was completely removed and now I have included it in submenu.

Adam-it:
> made a minor change here @adam, the deploy app catalog action was completely removed and now I have included it in submenu.

Thanks for the help and awesome find 👍

Issue #430, opened 2025-03-24
💡 [Feature]: Show `Upgrade` app when `install` action fails


Adam-it:
### 🎯 Aim of the feature

The idea is to check the error response of `install` action and when it failed due to the issue that the app was already installed on that site we could in the error message show and `upgrade` button which will run the upgrade action on that app.

https://github.com/pnp/vscode-viva/blob/3c8c45f9b3d2e8ada009cdd25e9b12080ec7b384/src/services/actions/CliActions.ts#L444-L454

It makes more sense to suggest the upgrade action to the user if the install action failed because the app is already there. In this case the user maybe would like to align and update the app to latest version which is what we may also do.

PR #468, opened 2025-05-04
Refactors SPFx actions. Closes #431, #430

### LINKED ISSUE #431: 💡 [Feature]: Show in progress kind of message after we click on any of the SPFx app actions
### 🎯 Aim of the feature

Currently when we click on any of the SPFx app actions the command is executed and VS Code show now response after clicking. We only get a success or error message when the action is done. 
The problem is that if we don't present any reaction like 'in progress' after we click on the action the user may think that something did not work and may click multiple times. This is what I did 😅.
The idea is to show loader after we click on any of those actions.

What we should do is `window.withProgress` and a message like `Deploying app... Check [output window](command:${Commands.showOutputChannel}) to follow the progress.`
For example other place were we already use it and we may do the same approach 
https://github.com/pnp/vscode-viva/blob/3c8c45f9b3d2e8ada009cdd25e9b12080ec7b384/src/services/actions/CliActions.ts#L1115-L1121

those are the action we should adapt

Image

_No response_

Adam-it:
## 🎯 Aim

The aim of this PR is to refactor SPFx actions to show progress when they are running and handle a special case when we try to install an app which is already installed to suggest either upgrade if it is the same app or remove if it is a different version of the app with the same version.

## 📷 Result

image

image

## ✅ What was done

- [X] Added show progress notify for all SPFx app actions
- [X] Moved SPFx app actions to separate file
- [X] Added handling when app installed or different app with same if installed path with quick remove or upgrade action in notify

## 🔗 Related issue

I know I know... one issue one PR, but I couldn't help myself to do it in a single change 😜

Closes: #431, #430


nicodecleyre (approved):
All good 😄 Thank you for your contribution

PR #481, opened 2025-05-14
Increase security in GitHub Workflows.

Adam-it:
## 🎯 Aim

The aim is to address security issues with workflows we used. 
Along the way I also added a condition so that the workflows will only run in the context of our repo and not on the forked repos as well.
docs -> https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idif

Since GitHub workflows run based on repo default branch we should merge this to `main` branch so it will resolve the security issues ASAP.

## ✅ What was done

- [X] Added permissions 
- [X] Added condition

### DIFF
--- a/.github/workflows/check-tests.yml
+++ b/.github/workflows/check-tests.yml
@@ -6,8 +6,12 @@ on:
 
 jobs:
   build:
+    if: github.repository == 'pnp/vscode-viva'
     name: "Test"
     runs-on: windows-latest
+
+    permissions:
+      contents: read
 
     steps:
       - name: Setup node
--- a/.github/workflows/check-tests.yml
+++ b/.github/workflows/check-tests.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build:
-    if: github.repository == 'pnp/vscode-viva'
+    if: github.repository_owner == 'pnp'
     name: "Test"
     runs-on: windows-latest
 
--- a/.github/workflows/release-beta.yml
+++ b/.github/workflows/release-beta.yml
@@ -5,9 +5,14 @@ on:
 
 jobs:
   build:
+    if: github.repository == 'pnp/vscode-viva'
     name: "Build and pre-release"
     runs-on: ubuntu-latest
 
+    permissions:
+      contents: write
+      packages: read
+    
     steps:
       - name: Setup node
         uses: actions/setup-node@v4
--- a/.github/workflows/release-beta.yml
+++ b/.github/workflows/release-beta.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   build:
-    if: github.repository == 'pnp/vscode-viva'
+    if: github.repository_owner == 'pnp'
     name: "Build and pre-release"
     runs-on: ubuntu-latest
 
--- a/.github/workflows/release-docs.yml
+++ b/.github/workflows/release-docs.yml
@@ -46,7 +46,7 @@ jobs:
           path: docs/dist
 
   deploy_docs:
-   if: github.repository_owner == 'pnp'
+    if: github.repository_owner == 'pnp'
     needs: build_docs
     runs-on: ubuntu-latest
 
--- a/.github/workflows/release-docs.yml
+++ b/.github/workflows/release-docs.yml
@@ -10,6 +10,7 @@ on:
 
 jobs:
   build_docs:
+    if: github.repository == 'pnp/vscode-viva'
     name: "Build and release docs"
     runs-on: ubuntu-latest
 
@@ -45,6 +46,7 @@ jobs:
           path: docs/dist
 
   deploy_docs:
+    if: github.repository == 'pnp/vscode-viva'
     needs: build_docs
     runs-on: ubuntu-latest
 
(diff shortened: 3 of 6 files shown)


.github/workflows/check-tests.yml
```diff
@@ -6,9 +6,13 @@ on:
 
 jobs:
   build:
+    if: github.repository == 'pnp/vscode-viva'
```
Saurabh7019, line 9: Should we check the owner instead? especially when we have talked about repo rename a couple of times? CLI has this - `if: github.repository_owner == 'pnp'`

Adam-it: good comment 👍 
Yes I will update that to your suggestion

Adam-it, line 9:
```suggestion
    if: github.repository_owner == 'pnp'
```

.github/workflows/release-beta.yml
```diff
@@ -5,9 +5,14 @@ on:
 
 jobs:
   build:
+    if: github.repository == 'pnp/vscode-viva'
```
Adam-it, line 8:
```suggestion
    if: github.repository_owner == 'pnp'
```

.github/workflows/release-docs.yml
```diff
@@ -10,6 +10,7 @@ on:
 
 jobs:
   build_docs:
+    if: github.repository == 'pnp/vscode-viva'
```
Adam-it, line 13:
```suggestion
    if: github.repository_owner == 'pnp'
```

.github/workflows/release-docs.yml
```diff
@@ -45,6 +46,7 @@ jobs:
           path: docs/dist
 
   deploy_docs:
+    if: github.repository == 'pnp/vscode-viva'
```
Adam-it, line 49:
```suggestion
   if: github.repository_owner == 'pnp'
```

.github/workflows/release-local.yml
```diff
@@ -5,9 +5,13 @@ on:
 
 jobs:
   build:
+    if: github.repository == 'pnp/vscode-viva'
```
Adam-it, line 8:
```suggestion
    if: github.repository_owner == 'pnp'
```

.github/workflows/release.yml
```diff
@@ -8,9 +8,13 @@ on:
 
 jobs:
   build:
+    if: github.repository == 'pnp/vscode-viva'
```
Adam-it, line 11:
```suggestion
    if: github.repository_owner == 'pnp'
```

.github/workflows/update-samples.yml
```diff
@@ -7,9 +7,13 @@ on:
 
 jobs:
   update:
+    if: github.repository == 'pnp/vscode-viva'
```
Adam-it, line 10:
```suggestion
    if: github.repository_owner == 'pnp'
```

Adam-it:
@Saurabh7019 I am ready for another review round 👍

.github/workflows/release-docs.yml
```diff
@@ -45,6 +46,7 @@ jobs:
           path: docs/dist
 
   deploy_docs:
+   if: github.repository_owner == 'pnp'
```
Saurabh7019, line 49: Indentation seems to be wrong here

.github/workflows/release.yml
```diff
@@ -8,9 +8,13 @@ on:
 
 jobs:
   build:
+    if: github.repository_owner == 'pnp'
     name: "Build and release"
     runs-on: ubuntu-latest
 
+    permissions:
+      contents: read
+
```
Saurabh7019, line 17: question - release.yml is almost same as rease-beta.yml. Shouldn't the permission be same as well?

Saurabh7019 (changes requested):
Just one minor change, something I could have fixed while merging, but I also had a question, so I thought I should get your thoughts first.

.github/workflows/release.yml
Adam-it: good point. TBH those are permissions proposed by GH security scanning and I did not question them 😉.
Lets align 👍

.github/workflows/release.yml
```diff
@@ -8,9 +8,13 @@ on:
 
 jobs:
   build:
+    if: github.repository_owner == 'pnp'
     name: "Build and release"
     runs-on: ubuntu-latest
 
+    permissions:
+      contents: read
```
Adam-it, line 16:
```suggestion
    permissions:
      contents: write
      packages: read
```

.github/workflows/release-docs.yml
```diff
@@ -45,6 +46,7 @@ jobs:
...
     needs: build_docs
     runs-on: ubuntu-latest
```
Adam-it, line 51:
```suggestion
  deploy_docs:
    if: github.repository_owner == 'pnp'
    needs: build_docs
    runs-on: ubuntu-latest
```

.github/workflows/release-docs.yml
Adam-it: sorry for that. I was updating it using web UI as review comments. I will try again

Saurabh7019:
Merged. Thank you, Adam. You rock :)

Issue #435, opened 2025-03-29
💡 [Feature]: Add more manage capabilities for Tenant Wide extensions


Adam-it:
### 🎯 Aim of the feature

Currently, the extension just shows the tenant-wide extensions on our SPO tenant. The idea is to add some management capabilities to it. Some ideas from the top of my head are:
- Remove 
- Enable/Disable 

### 📷 Images (if possible) with expected result

![Image](https://github.com/user-attachments/assets/8e0d2d75-f7e7-4409-a756-23704c3f607c)

### 🤔 Additional remarks or comments

Possible things we could allow to modify 
https://learn.microsoft.com/en-us/sharepoint/dev/spfx/extensions/basics/tenant-wide-deployment-extensions

Adam-it:
@Saurabh7019, @nicodecleyre I would appreciate your feed on this feature 🙏. 
What do you think?

Saurabh7019:
sounds great; it will only make the tool richer. I will be happy to implement it when it's open.

Adam-it:
> sounds great; it will only make the tool richer. I will be happy to implement it when it's open.

Thanks for the feed. 
What do you think if the open points? Like the ability to modify list template or web template? The problem here is that those actions would not make sense for every extension type. 
Also we could allow to modify the order/priority. This number specifies the order in which the extensions are loaded on a page.
What do you think?

Saurabh7019:
Users can update Title, Properties, Web template, List template, Sequence, etc. from editing the item in the Tenant Wide Extensions list, and I think we should also support it from the SPFx TK without worrying if updates have any effect on the extension or not.

I would think of actions like - 
- Rename
- Update Properties
- Disable
- Enable
- Update Sequence
- Update Host Properties

too much? maybe, we can start with the ones you initially mentioned.

nicodecleyre:
> > Users can update Title, Properties, Web template, List template, Sequence, etc. from editing the item in the Tenant Wide Extensions list, and I think we should also support it from the SPFx TK without worrying if updates have any effect on the extension or not.
> > I would think of actions like -
> > 
> > * Rename
> > * Update Properties
> > * Disable
> > * Enable
> > * Update Sequence
> > * Update Host Properties
> > 
> > too much? maybe, we can start with the ones you initially mentioned.
> 
> I like your proposal. Let's do it like this. Let's leave the scope of this issue as is and do a follow up issue with the additional stuff you mentioned 👍
> 
[...]

PR #529, opened 2025-06-25
Issue 435

Saurabh7019:
## 🎯 Aim

> This PR adds new management capabilities for Tenant Wide Extensions in the SPFx Toolkit

## 📷 Result

image

## ✅ What was done

- [X] Added support to enable or disable Tenant Wide Extensions from the app catalog tree
- [X] Added support to remove Tenant Wide Extensions
- [ ] Removing a Tenant Wide Extension does **not** remove the app as the extension may be part of a larger solution with other components. However, removing an app does remove the associated tenant-wide extension (if any) 

## 🔗 Related issue

Closes: #435

### DIFF
--- a/package.json
+++ b/package.json
@@ -840,6 +840,30 @@
 				"title": "Remove Site App Catalog",
 				"category": "SPFx Toolkit",
 				"icon": "$(trash)"
+			},
+			{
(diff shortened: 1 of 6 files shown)


Adam-it (approved):
LGTM ✅
tested locally ✅

PR #544, opened 2025-07-25
Add update action to tenant wide extensions management capabilities. Closes #439

### LINKED ISSUE #439: 💡 [Feature]: Add more actions to tenant wide extensions management capabilities
### 🎯 Aim of the feature

The idea is to allow to update Title, Properties, Web template, List template, Sequence, etc. the same way like it is currently possible by editing the item in the Tenant Wide Extensions list.

We could try to introduce the following actions:
- Rename
- Update Sequence
- Update Web Template
- Update List template
- Update Host Properties

_No response_

_No response_

Saurabh7019:
## 🎯 Aim

This PR adds update action to tenant-wide extensions management capabilities.

## 📷 Result

image

## ✅ What was done

- [X] Added ability to update multiple tenant-wide extension properties in a single action including Title, Component Properties, Web Template, List Template, Sequence, and Host Properties
- [X] Shows curretn values for each properties and input boxes populate the existing values for editing
- [X] Updated `getTenantWideExtensions` to return complete extension object including all properties

## 🔗 Related issue

Closes: #439


Adam-it (approved):
Awesome 👏👏👏 
I am generally not sure about how we show the summary in the field placeholder. 
When we have a looot of fields to update it is totally not readable, But on the other hand I don't have any easy alterative for now. 
Let's merge since this is not a killer blocker just more a thing of preference I guess

PR #575, opened 2025-09-07
Adds CLI for Microsoft 365 package clean up

Adam-it:
## 🎯 Aim

The aim is to perform clean up of CLI for Microsoft 365 npm package not to include any docs, spec.js and js.map files in the final vs code extension bundle 

## ✅ What was done

- [X] updated CLI for Microsoft 365 copy PowerShell script


Adam-it (commented):
checked

Issue #586, opened 2025-09-27
💡 [Feature]: Update spo_page_add Language Model Tool to follow standards


Saurabh7019:
### 🎯 Aim of the feature

Refactor the existing spo_page_add Language Model Tool to align with updated naming conventions, improve model descriptions, fix parameter mismatches, and add missing functionality options.

https://github.com/pnp/vscode-viva/blob/d0e5890c76e0d1de24ee9fa9cbb6fe8098cde95c/package.json#L224-L260

**Issues to address:**

- Rename the tool to **add_spo_page** following the naming convention {verb}_{noun} as established in the codebase
- Improve modelDescription:
  - What exactly does the tool do? - describe the specific functionality and capabilities
  - What kind of information does it return? - detail the expected response format
  - When should and shouldn't it be used? - provide clear use cases and scenarios to avoid
  - Describe important limitations or constraints of the tool - identify technical and permission limitations
- Refactor [PageAdd](https://github.com/pnp/vscode-viva/blob/dev/src/chat/tools/spo/PageAdd.ts)
  - Parameter mismatch: Interface uses `baseTemplate` but `package.json` defines `layoutType`
  - Missing properties: The `package.json` defines `name`, but the interface doesn't have it. Let's also include additional CLI options `description`, `promoteAs`, and `commentsEnabled`
 
Under the hood use: [spo page add](https://pnp.github.io/cli-microsoft365/cmd/spo/page/page-add)

### 🤔 Additional remarks or comments

- VS Code LM docs: [https://code.visualstudio.com/api/extension-guides/tools](https://code.visualstudio.com/api/extension-guides/tools)
- CLI for Microsoft 365 Reference: [https://pnp.github.io/cli-microsoft365/cmd/spo/page/page-add](https://pnp.github.io/cli-microsoft365/cmd/spo/page/page-add)

Adam-it:
@Saurabh7019, exactly what we need to get done. 
As I mentioned, this is something we need to do first and get merged before we proceed with other issues related to this epic. 
IMO this way we will ensure the new LM Tools in this area will follow the correct standard as most probably new tools will use this one as reference

PR #596, opened 2025-09-29
Updates spo_page_add Language Model Tool. Closes:  #586

Adam-it:
## 🎯 Aim

The aim is to update the `spo_page_add` LM Tool to follow the new standard, update this LM Tool's input parameters, and add some new ones that were missing

## 📷 Result

As part of this issue, I moved all LM Tools to subcatalogs, as most probably in the future we will have a lot of thos,e some tried to make it a more granular structure
image

for the following prompt 
image

I got this result
image

## ✅ What was done

- [X] Moved all language model tools to sub-catalog structure 
- [X] Renamed page add LM tool to `add_spo_page`
- [X] Updated its input properties and added others that were missing

## 🔗 Related issue

Closes: #586


Adam-it:
> Hi @Adam, 
> I noticed, the tool creates an empty new article page as expected, but as soon as I click the 'Edit' button on the page, the page suddenly gets populated with content, and the CanvasContent1 field changes from null to containing actual content. Do you observe the same behavior?

Will need to double check during night time.
Indeed very strange behavior. I wonder what will happen if we just use the CLI command from command line to do the same thing 🤔. Do we also have the same strange result, if so, then is it only done by our version of CLI or also happens with the officials CLI repo. 
I believe this is something that would need to be fixed in the CLI command level but just need to understand if it will be the official PnP one or the our commonJS version of it.

Adam-it:
> the issue is with our version of CLI.

aaa ok. Thanks for the investigation. 
The fix should be quite easy as most probably there was some fixup done in the official CLI version which I did not align with our version of CLI.
This is quite of a problem and I have it on my list to do a global alignment of all SPO related commands between official CLI version and out version of CLI but due to the ESM lack of support I will need to do it a bit manually as I may not depend on the prompt npm package or ZOD etc.

Anyway I will do the alignment of this one `spo page add` command and I think it should fix it.
I will try to do it today overnight and let you know so we will proceed with merging this PR when this will get fixed on the CLI in commonJS side.

Saurabh7019:
I included the fix in the cli-cjs branch locally and it works fine now. All the other changes in this PR look perfect. Should I am merge this PR since the cli-cjs updates are not directly impacting it?

Adam-it:
@Saurabh7019 nevermind. Also got it fixed just now. I am doing final test and should merge it to cjs version in like 30 min

Adam-it:
yep, rechecked and works
image

Adam-it:
@Saurabh7019 I just pushed the fix on the cjs version of cli so we may do a final recheck on your side and proceed with the merge IMO.
BTW I really need to prioritize the alignment between what we have done in the official PnP CLI npm package and our cjs version of it. Again time is the main factor here 😕 as it is not just a simple merge of the SPO folder between those packages as I would need to revert all ESM package/code to CJS which means refactoring/reverting the ZOD updates in those commands and new prompt library, and some other small things like removing `.js` in the imports etc.

it would reeeelly help if VS Code would add full support for extension with ESM dependencies. Now it is getting better but still in many parts not possible like in extensions with webViews 🤨

Saurabh7019:
build error in cli-cjs branch, can you remove the extra parameter before we pre-release?

image

Adam-it:
this is strange... something gonne wrong when I merged the changes 😕

Adam-it:
@Saurabh7019 I committed a fixup and did a quick recheck.
Not sure what happened with my last commit 🤔

PR #626, opened 2025-10-28
Adds Heft tasks. Closes #618

### LINKED ISSUE #618: 💡 [Feature]: Show `heft` tasks as part of Tasks view
### 🎯 Aim of the feature

The aim is to show most relevant for SPFx development heft cli tasks in the Tasks view.
The heft tasks should be shown in the same list as gulp tasks but in a separate tree branch dedicated for heft. 
Currently we should have already two branches in this tree view: first for gulp and second for npm scripts.

New heft tasks should only be present for newer projects of SPFx (after 1.22 release) that are note ejected and use heft as primary build toolchain tool. Most probably it will replace gulp tasks in the tree view. 

_No response_

follow up to #424 we should wait for the related issue to get merged/completed before implementing this issue 
also lets wait for SPFx 1.22 to get a RC version to check based on which properties in the SPFx project would be best to determine if the heft tasks should be visible for SPFx project or not.

Adam-it:
## 🏗️ ~~Work in progress~~

~~This is still work in progress and we may not merge it just yet as this was developed based on SPFx 1.22-beta.4 so still some things may change before the official GA, but this PR should give us a head start when it comes to adding heft tasks support to this extension~~

## 🎯 Aim

The aim is to refactor the task view to show `heft` tasks in case the SPFx project uses primary heft for build toolchain tasks

## 📷 Result

SPFx 1.22 with Heft

image
image

SPFx 1.21 with Gulp

image
image

## ✅ What was done

- [X] Refactored the icons and naming a bit
- [X] Added heft tasks to the task view in case the project uses heft npm package
- [x] Update Task View tests to support heft
- [x] Optimize code, currently most is duplicated but it seems many methods may be reused and we only need to pass if `gulp` or `heft` should be used in the command
- [x] It seems `heft` does not support `start` (previously serve) with a specific configuration, like for command sets. Left a TODO for it in code and checking this with SPFx team

## 🔗 Related issue

Closes: #618

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ Sign-in is also required for some actions to work properly like the deploy actio
 
 ### 7️⃣ Tasks
 
-The extension shows all possible tasks one may run on an SPFx project. The tasks view will show all relevan tasks that are available for the current SPFx project and all npm scripts that are defined in the package.json file.
+The extension shows all possible tasks that can be run on an SPFx project. The tasks view displays all relevant tasks available for the current SPFx project, as well as all npm scripts defined in the `package.json` file.
 
 For SPFx projects created with versions below 1.22, Gulp tasks will be shown in the tasks view:
 
--- a/assets/walkthrough/tasks.md
+++ b/assets/walkthrough/tasks.md
@@ -18,20 +18,20 @@ The task view will also show all npm scripts that are defined in your package.js
 
 ![Tasks](../images/tasks.png)
 
-From 1.22 version onwards, SharePoint Framework uses [Heft](https://heft.rushstack.io/) as the pluggable build system for developing for the SharePoint Framework.
+From version 1.22 onwards, SharePoint Framework uses [Heft](https://heft.rushstack.io/) as the pluggable build system for developing with the SharePoint Framework.
 
-SPFx Toolkit will continue to support Gulp tasks for existing projects, but for new projects created with SPFx 1.22 and above, Heft tasks will be shown in the tasks view along side npm scripts.
+SPFx Toolkit will continue to support Gulp tasks for existing projects, but for new projects created with SPFx 1.22 and above, Heft tasks will be shown in the tasks view alongside npm scripts.
 
 The view allows you to run all Heft tasks that are supported in SPFx projects:
 
 - build - Builds the client-side solution project.
-- clean - Cleans the client-side solution project's build artifacts from the previous build and from the build target directories (lib and dist).
+- clean - Cleans the client-side solution project's build artifacts from the previous builds and from the build target directories (lib and dist).
 - start - Serves the client-side solution project and assets from the local machine.
 - test - Runs unit tests, if available, for the client-side solution project.
 - package-solution - Packages the client-side solution into a SharePoint package.
 - deploy-azure-storage - Deploys client-side solution project assets to Azure Storage.
 
-and more...
+and more.
 
 ![Tasks](../images/heft-tasks.png)
 
--- a/docs/src/content/docs/features/coding.mdx
+++ b/docs/src/content/docs/features/coding.mdx
@@ -9,11 +9,11 @@ SPFx Toolkit comes with a set of features that will speed up your development pr
 
 ### Heft
 
-For newer version of SPFx (1.22 and above), Heft is used as the pluggable build system for developing for the SharePoint Framework.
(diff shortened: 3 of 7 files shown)


Adam-it:
@pnp/spfx-toolkit-maintainers I am wondering if this PR and functionality is actually needed. I am afraid it will create more confusion and I will try to explain why. 
In the past (SPFx 1.21 and earlier) we needed to have `gulp-cli` installed as one of the global dependencies and Gulp locally (in project) is because of the incompatibility of Gulp v3 & v4. The Gulp CLI, installed globally, could abstract its incompatibility so you could use gulp at the command line anywhere (https://www.voitanos.io/blog/mea-culpa-always-install-gulp-cli-globally-not-gulp/)

But now it is not the case anymore. For SPFx 1.22 which now will use heft we do not need `heft` installed globally anymore and we may just use the one which is part of the SPFx project installed locally using `./node_modules/.bin/heft ... ` like `./node_modules/.bin/heft build`. 

To prove this I was managed to build a SPFx 1.22 project using `npm run build` which runs `heft build --clean` withouth having `heft` installed locally. 
When I just run `heft build` it fails as I do not have the package
image

SO, I believe this PR may actually bring the following problems:

- heft tasks we expose as part of this PR will only work if you have `heft` installed globally which is not needed
- the heft tasks mostly duplicate the once exposed as part of `npm run` command so it clutters the view. If we just leave `npm` tree for the view will be smaller and still have everything that is needed for the SPFx project

for older projects we still need to show both `npm` tree (which is kinda useless for older projects. and `gulp` which expose `gulp-cli` tasks. So in this case we should not change anything.

What do you think? After more tests of 1.22 and investigating around I suggest NOT to merge this PR and leave only npm scripts tree in our task view for project that are SPFx 1.22 and newer versions

Adam-it:
So I did some more testing of SPFx 1.22 and I think that `heft` is indeed needed. For example, one problem is that with just npm run build or package-solution you may not build the package for production, you still need heft to do that.

So now I think this PR might be worth adding and also installing heft globally as well in the install action.
@pnp/spfx-toolkit-maintainers what do you think?

Saurabh7019:
> The PR which was merged a few weeks ago already introduced a tree view in tasks view having both gulp and npm scripts

This is exactly what we want.. keep both Gulp Tasks (rename it to SPFx Tasks) and NPM Scripts, means no UI changes. For older versions, everything will continue to work as before. For newer versions, we can update the functions below to run the Heft tasks as well.

- cleanProject
- buildProject
- testProject
- trustDevCert
- deployToAzureStorage
- serveProject
- bundleProject
- packageProject
- publishProject
- cleanProject

Saurabh7019:
> what do you mean by 'core'? for some its just build and package-soltion,.

for me, it is clean, bundle, package, serve, test
but I agree, filtering out what we think are the core tasks might not be a good idea. We should just keep the nodes collapsed.

assets/walkthrough/actions.md
```diff
@@ -18,9 +18,9 @@ And for Azure DevOps👇
 
 ![Azure DevOps CI CD pipeline](../images/azdo-ci-cd.gif)
 
-[Check out our docs for more details](https://pnp.github.io/vscode-viva/features/actions/#cicd-workflow)
+[Check out our docs for more details](https://pnp.github.io/vscode-viva/features/ci-cd/)
 
-- **Upgrade project** - Uses CLI for Microsoft 365 to create a .md and/or code tour report with upgrade guidance to the latest supported SPFx version by the extension. The upgrade guidance provides optional as well as required steps a user needs to take manually in order to upgrade the project. At the end of the guidance file, the user may find a summary of the steps in a single script and a couple of manual steps that need to be taken.
+- **Upgrade project** - Uses CLI for Microsoft 365 to create a .md and/or code tour report with upgrade guidance to the selected SPFx version. The upgrade guidance provides optional as well as required steps a user needs to take manually in order to upgrade the project. At the end of the guidance file, the user may find a summary of the steps in a single script and a couple of manual steps that need to be taken.
```
nicodecleyre: Let's try to rephrase this sentence - especially the part ´as well as required steps´ breaks the flow. Proposal: ´**Upgrade project** - Generates a detailed upgrade report (.md and/or code tour) to guide users from their current SPFx version to a selected target version using CLI for Microsoft 365. The report includes both required and optional upgrade steps, with a consolidated script at the end for automated actions and a list of remaining manual steps.´

PR #629, opened 2025-11-02
Refactor the way we add CLI for Microsoft 365 to SPFx Toolkit. Closes #526

### LINKED ISSUE #526: Update the way we include CLI for Microsoft 365 in our project
As of now, VS Code extensions still don't fully support esm package #114.
So to go around this problem we maintain CLI for Microsoft 365 in cjs and include it to this extension with a [hack](https://github.com/pnp/vscode-viva/blob/main/contributing.md#-how-to-start---minimal-path-to-awesome) to build CLI for Microsoft 365 locally and then use a script to replace the dependencies in the node_modules folder.

This approach was fine but it looks like the esm support is not something that will be fixed quite quick and also we will need to maintain some package for CLI for Microsoft 365 anyway we could update this process. 
Instead of doing our hack what we could do is:
1. update the package.json of our version of CLI for Microsoft 365 maintained in cjs to change the package name
2. publish it to the [npm](https://www.npmjs.com/) but making it also super clear in the readme that this package should only be used as part of SPFx Toolkit and the pnp CLI for Microsoft 365 should be used otherwise
3. in our package.json we should reference our new package instead of the pnp one

update docs and contributing guide as well

Adam-it:
## 🎯 Aim

The aim is to refactor SPFx Toolkit to use our own version of CLI for Microsoft 365 npm package instead of the hacky current approach. As a result, I published our own version of CLI for Microsoft 365 for SPFx Toolkit (https://www.npmjs.com/package/@pnp/cli-microsoft365-spfx-toolkit) in CommonJS, and now SPFx Toolkit simply uses this package instead of doing it manually. This simplified development also allows us to have a more formal version of CLI for Microsoft 365 that is based on the official package but has some tweaks created for this project

## 🧪 Ideas to test

I would suggest to run `npm install` and `vsce package` to create the local version of the extension locally and install it in VS Code and click around the SPFx Toolkit mainly focusing on features that use CLI for Microsoft 365 commands.

## ✅ What was done

- [X] Updated npm package.json to use CLI for Microsoft 365 SPFx Toolkit package and updated all imports
- [X] deleted the hacky script to ingest CLI for Microsoft 365 from the local repo build
- [X] cleaned up contributing guidance
- [X] cleaned up GitHub Workflows not to build CLI for Microsoft 365 repo and added the files manually

## 🔗 Related issue

Closes: #526


Adam-it:
ok one point I missed is that we should get rid of `fsevents` before we proceed

Adam-it:
Setting this to draft until I solve the `fsevents` issue

Adam-it:
Ready for review 🤩

nicodecleyre:
Thank you for your PR 👏👏👏 ! We'll review is ASAP

nicodecleyre (approved):
Tested locally, works like a charm! Awesome work @Adam-it 👏👏👏

PR #638, opened 2025-11-09
Adds support for all SPFx versions to upgrade action and LM tool. Closes #517

### LINKED ISSUE #517: 💡 [Feature]: Refactor Upgrade project SPFx version feature to support multiple SPFx versions
### 🎯 Aim of the feature

~~'Upgrade project SPFx version' - So what we could do is first we could run and check if there is any SPFx-generator installed and get its version. If there isn't then we could do what we do now which is providing guidance to upgrade to latest version of SPFx. But if we will find some SPFx-generator we could run the `m365 spfx project upgrade with the --toVersion` option and specify the version we found in the globally installed SPFx-generator.~~

Since the PR https://github.com/pnp/vscode-viva/pull/621 introduces SPFx Compatibility Matrix with all SPFx versions I suggest that when someone would click on SPFx project upgrade we present the list of SPFx versions that are higher (later) than the version of SPFx in the current project and allow the user to pick any version of SPFx and run upgrade to it.
If someone would just click enter without selecting any version of SPFx from the list we would run upgrade version to latest version of SPFx.

_No response_

_No response_

Adam-it:
## 🎯 Aim

The aim is to refactor the upgrade action and LM tool to allow to specify any SPFx version to which we want to upgrade to which is higher than the SPFx version of the current project

## 📷 Result

When the current SPFx project is in the latest version of SPFx based on the SPFx Compatilbity Matrix we get this information 
image

When selecting an upgrade action, we get a picker with versions higher than current one 
image

It is possible to get upgrade report for the specified version
image

Also updates upgrade LM tool that now also picks up version from conversation
image

If not specified will generate report to the latest version of SPFx
image

## ✅ What was done

- [X] updated upgrade action to show picker with SPFx version higher than the current one. defaults to the latest version of SPFx
- [X] updated upgrade LM tool to pick up the SPFx version, validate it with the compatibility matrix and use it, in case it is not specified, upgrade to latest version
**- [ ] Update Readme and docs ---> //TODO: work in progress 🏗️**

## 🔗 Related issue

Closes: #517

### DIFF
--- a/package.json
+++ b/package.json
@@ -552,7 +552,16 @@
 				"userDescription": "Upgrade SharePoint Framework project.",
 				"canBeReferencedInPrompt": true,
 				"icon": "$(arrow-up)",
-				"inputSchema": {}
+				"inputSchema": {
+					"type": "object",
+					"properties": {
+						"toVersion": {
+							"type": "string",
+							"description": "Target version to upgrade the SharePoint Framework project to. This option is optional.",
+							"default": ""
+						}
+					}
+				}
 			}
 		],
 		"walkthroughs": [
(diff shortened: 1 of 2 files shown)


nicodecleyre:
Thank you for your contribution 👏💪. We'll review ASAP

Issue #683, opened 2026-01-06
🐞 Bug report: broken image link in readme


Adam-it:
### ⭐ Priority

(Low)☹️ Something is a little off

### 📝 Describe the bug

It seems the install dependencies image link is broken in the readme file

https://github.com/pnp/vscode-viva/blob/2ee124aeeff1041661624d6d8414fdf6322fb558/README.md?plain=1#L75

### 👣 Steps To Reproduce

read the readme file 😁

### 📜 Expected behavior

should see the image

### 📷 Screenshots

<img width="1008" height="420" alt="Image" src="https://github.com/user-attachments/assets/e68a8c71-c1d7-49e9-94da-1f1a96735bf8" />

Adam-it:
> I can give this a go!

All yours 👍

PR #685, opened 2026-01-09
Resolve missing install-dependencies image path

nirav-raval:
## 🎯 Aim

Resolve install dependencies image link which is broken in the readme file

## 📷 Result

 **Before:**

image

**After:**

image

## ✅ What was done

The image path was incorrect. Added proper url which is now pointing towards the correct image.

- [X] Updated image path to correct image location

## 🔗 Related issue

Closes: #683

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ This action checks if your local development environment meets the required vers
 
 It is also possible to install the required Node.js version and global dependencies for any SPFx version using the **Install dependencies** action.
 
-![install dependencies](./assets/images/install-dependencies.png)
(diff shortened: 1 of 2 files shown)


Saurabh7019:
Thanks for this fix, @nirav-raval. I just noticed that the same issue also exists in [https://github.com/pnp/vscode-viva/blob/dev/assets/walkthrough/validate-local-setup.md](https://github.com/pnp/vscode-viva/blob/dev/assets/walkthrough/validate-local-setup.md)

As part of this PR, could you also please update all image references in that file to use the correct paths?

PR #689, opened 2026-01-13
Allow bulk operations for all apps in an app catalog, Closes #442

### LINKED ISSUE #442: 💡 [Feature]: Allow bulk operations for all apps in an app catalog
### 🎯 Aim of the feature

The aim is to allow the same operations/actions that are available for each app to be executed on app catalog level and therefore when some actions is used it will be performed for every app in that app catalog.
The aim is to allow the following actions:
- Retract
- Remove
- Enable
- Disable
- Install
- Uninstall
- Upgrade

This is based on a real case scenario where today I needed to perform a clean up of all apps I had added to a site from a site level app catalog and I needed to perform it one by one. 

The idea is to allow those operations on app catalog level 

Image

_No response_

nicodecleyre:
## 🎯 Aim

This issue allows bulk operations for all apps in an app catalog

## 📷 Result

Screenshot 2026-01-14 003404

## ✅ What was done

- [X] bulk operations to:
- Deploy
- Retract
- Remove
- Enable
- Disable
- Install
- Uninstall
- Upgrade

## 🔗 Related issue

Closes: #442


Adam-it (changes requested):
@nicodecleyre awesome work!
I played around with it a bit and checked most of the bulk options

Image

What we forgot in this PR is updating the docs. Since docs are now part of our code we should update it together with the PR so when it gets shipped we wont miss to align the documenatation. 

Also I would move the buld operations to a new `SpgxAppCLIBulkAction.ts` file to make for better maintability so we dont make the SpgxAppCLIAction to long

docs/src/content/docs/features/management-capabilities.mdx
```diff
@@ -46,3 +46,20 @@ You can expand each app catalog to reveal the list of apps contained within. Sel
 - **Move**: Moves the solution to a different app catalog.
 
 ![app-catalog-list](../../../../assets/images-vscode/app-catalog-list.png)
+
+### Bulk app operations
```
Adam-it: I think it would be cool if we had at least one screenshot of this feature.
Keep in mind we should always use default visual studio light theme for that

src/services/actions/SpfxAppCLIBulkActions.ts
```diff
@@ -0,0 +1,732 @@
...
+                    }
+                }
+
+                progress.report({ message: `Completed: ${completed} of ${apps.length} app(s) uninstalled.` });
+            });
+
+            Notifications.info(`Bulk uninstall completed: ${apps.length} app(s) processed.`);
+
+            // refresh the environmentTreeView
+            await commands.executeCommand('spfx-toolkit.refreshAppCatalogTreeView');
```
Adam-it: This code is very similar to what we have in lines 454-531. 
I think it would be best if we could somehow extract it to a reusable private method

Adam-it (changes requested):
@nicodecleyre ready for another round? 
I rechecked the PR and left a few comments we could recheck. Let me know if you need any help

Adam-it (changes requested):
@nicodecleyre I added a few fixups but the PR is not ready yet.
I think during the refactor in the PR there was a logical error introduced in which in many places the url to site level app catalog is introduced instead of the url to the site 

Image

I am still investigating this but from what I tested most of the bulk operations won't work due to this error. Maybe you wanna check this case as well. 

Also shouldn't those bulk operations be also visible on tenant level app catalog? 
I think it should that is why in SpfxAppCLIBulActions.ts in lines 603-618 we handle the case in which the app catalog is null and then the user may specify the relative site URL to which the app should be installed/uninstalled 🤔

Adam-it:
we also need to rebase this branch against latest dev since we just had a minor release

Adam-it:
We just did a minor release of SPFx Toolkit, agian. Please rebase this branch to align with latest dev before we proceed

Adam-it:
@nicodecleyre we had another minor release and we need to rebase this one again.
Also this PR is open for a long time and is quite inactive. Do you think it still makes sense? Maybe if we do not have time to commit to this feature lets close the PR and reopen the issue?

nicodecleyre:
> @nicodecleyre we had another minor release and we need to rebase this one again. Also this PR is open for a long time and is quite inactive. Do you think it still makes sense? Maybe if we do not have time to commit to this feature lets close the PR and reopen the issue?

Ready

package.json
Adam-it, line 1372: I think we should add the buld operation commands here as well so that they do not show in the command pallete

Image

src/constants/ContextKeys.ts
```diff
@@ -18,5 +18,12 @@ export const ContextKeys = {
...
   copyApp: 'pnp.etv.app.copy',
-  moveApp: 'pnp.etv.app.move'
+  moveApp: 'pnp.etv.app.move',
+  bulkRetractApps: 'pnp.etv.bulk.retract',
+  bulkRemoveApps: 'pnp.etv.bulk.remove',
+  bulkEnableApps: 'pnp.etv.bulk.enable',
+  bulkDisableApps: 'pnp.etv.bulk.disable',
+  bulkInstallApps: 'pnp.etv.bulk.install',
+  bulkUninstallApps: 'pnp.etv.bulk.uninstall',
+  bulkUpgradeApps: 'pnp.etv.bulk.upgrade'
```
Adam-it, line 28: I wonder if this is needed. I mean it is not used anywhere in code right?

src/services/actions/SpfxAppCLIBulkActions.ts
```diff
@@ -0,0 +1,683 @@
...
+                            ...(appCatalogUrl && {
+                                appCatalogScope: 'sitecollection',
+                                appCatalogUrl: appCatalogUrl
+                            })
+                        };
+
+                        await CliExecuter.execute('spo app retract', 'json', commandOptions);
+                        completed++;
+                    } catch (e: any) {
+                        Notifications.error(`Failed to retract '${app.Title}': ${e?.error?.message || e?.message}`);
```
Adam-it, line 174: I wonder if this won't be a pain for the user. I mean if we have 30 apps and all 30 will fail the user in VS Code will get 30 notifies 🤔.
I think I more like what we do in the `bulkDeployAppCatalogApps` where in the loop we only add to logs if something fail and after everything is done we present a notification summary how many succeed and which fail

This comment actually applies to all other bulk app catalog commands below

src/services/actions/SpfxAppCLIBulkActions.ts
```diff
@@ -0,0 +1,683 @@
...
+                    }
+                }
+
+                progress.report({ message: `Completed: ${completed} of ${deployedApps.length} app(s) retracted.` });
+            });
+
+            Notifications.info(`Bulk retract completed: ${deployedApps.length} app(s) processed.`);
```
Adam-it, line 181: I think it would be best to do what we did `bulkDeployAppCatalogApps` so that we notify how many runs were correct and how many not. This comment also applies to all other bulk actions

src/services/actions/SpfxAppCLIBulkActions.ts
```diff
@@ -0,0 +1,683 @@
...
+     * Upgrades all apps on a site.
+     *
+     * @param node The tree item representing the app catalog.
+     */
+    public static async bulkUpgradeAppCatalogApps(node: ActionTreeItem) {
+        try {
+            const appCatalogUrl = SpfxAppCLIBulkActions.getAppCatalogUrlFromNode(node);
+            const isTenantCatalog = !appCatalogUrl || (EnvironmentInformation.appCatalogUrls && appCatalogUrl === EnvironmentInformation.appCatalogUrls[0]);
+
+            const apps = await CliActions.getAppCatalogApps();
```
Adam-it, line 470: shouldn't we also add `appCatalogUrl` param here? If it is not set won't it update the tenant level apps?

src/services/actions/SpfxAppCLIBulkActions.ts
```diff
@@ -0,0 +1,683 @@
...
+            await commands.executeCommand('spfx-toolkit.refreshAppCatalogTreeView');
+        } catch (e: any) {
+            const message = e?.error?.message || e?.message;
+            Notifications.error(message);
+        }
+    }
+
+    /**
+     * Installs all apps from the specified app catalog to a site.
+     * If installing to a site collection app catalog, apps will first be copied from tenant catalog.
```
Adam-it, line 381: don't quite get this comment 🤔. How tenant app catalog is used here?

Adam-it (changes requested):
@nicodecleyre unfortunately I still noticed a few issues with the PR a few of them quite important we need to address before we merge. Let's give this another go

PR #694, opened 2026-01-16
Updates scaffolding form. Closes #667

### LINKED ISSUE #667: 💡 [Feature]: Add notify that scaffolding is limited only to latest version of SPFx
### 🎯 Aim of the feature

As a follow up to this commnet https://github.com/pnp/vscode-viva/issues/514#issuecomment-3414463116
The aim is to add a warning message to tha scaffolding form, and to the `/new` chat participant command to that will inform the user that the scaffolding only works towards latest version of SPFx. 
We should show this warning message only after we detect the current version of SPFx the user has locally is not the latest. We should perform this check on every scaffolding form start (co after clicking on the action)

_No response_

_No response_

Adam-it:
## 🎯 Aim

The scaffolding form is not adapted for all versions of SPFx, and most probably it will take a lot of time until we do that. For now the aim is to adapt the scaffolding options in order to inform the user that they only support latest version of SPFx

## 📷 Result

image

## ✅ What was done

- [X] Added notification to the scaffolding form
- [X] Removed gulp fast serve
- [X] Updated chat participant prompts

## 🔗 Related issue

Closes #667

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ Install additional dependencies with a single click straight from the scaffoldin
 
 ***Important!***
 
-As of now creating new project supports only latest version of SharePoint Framework (SPFx). If you need to create a project with an older version please use the the SharePoint Yeoman generator directly in the terminal.
+As of now, creating new project supports only latest version of SharePoint Framework (SPFx). If you need to create a project with an older version, please use the the SharePoint Yeoman generator directly in the terminal.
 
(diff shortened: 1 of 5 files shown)


Adam-it:
// TODO

I just noticed I also need to update the images we have in docs in order to remove the fast serve which we do not allow anymore as it does not make sense for latest version of SPFx

Adam-it:
We just had a new minor release. We should rebase this branch before we proceed

PR #697, opened 2026-01-18
Adds validate local setup for current project action. Closes #518

### LINKED ISSUE #518: 💡 [Feature]: Add Validate local setup for current project action
### 🎯 Aim of the feature

Validate local setup for current project - we may think of better name but what I was thinking is to allow an action that will run the m365 spfx doctor command in the context of the currently opened SPFx project in VS Code. In this case the CLI command check the SPFx version that is used in the project based on the .yo-rc.json file and validates if the globally installed dependencies like gulp-cli and Node version are correct for that version of SPFx 🤯. I think it is an awesome feature for folks that will open some older SPFx project and would like to quickly check if they are using correct version of Node.js and gulp-cli for that project.

This should be visible as a new action in the action view 

Image

related CLI for Microsoft 365 command we should use under the hood in the context of the SPFx project [m365 spfx project doctor](https://pnp.github.io/cli-microsoft365/cmd/spfx/project/project-doctor)

Adam-it:
## 🎯 Aim

The aim is to add a new action that will validate user local setup for the currently open SPFx project. 
For example someone opens some old SPFx sample that uses SPFx 1.16 and would like to recheck if the current version of node and local dependencies are valid for that project. The aciton uses `spfx doctor` command to get the report and similar like the setup action will install the correct version of node and global dependencies if needed based on the compatibility matrix

## 📷 Result

image

image

## ✅ What was done

- [X] Added new action
- [ ] Updated docs 👉// TODO

## 🔗 Related issue

Closes: #518

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ And the code tour guidance will provide you upgrade tips directly in your code!
 
 [Check out our docs for more details](https://pnp.github.io/vscode-viva/features/actions#rename-current-project)
 
-- **Increase project version** - This action allows you to increase project major, minor or patch varsion and automatically align the versioning between `package.json` and `package-solution.json` files.
+- **Increase project version** - This action allows you to increase project major, minor or patch version and automatically align the versioning between `package.json` and `package-solution.json` files.
 
 ![Increase project version](./docs/assets/images-vscode/increase-versioning.png)
 
@@ -253,7 +253,7 @@ And the code tour guidance will provide you upgrade tips directly in your code!
 
 [Check out our docs for more details](https://pnp.github.io/vscode-viva/features/actions#deploy-project)
 
-- **Set Form Customizer** - Allows to update the New, Edit or View from of any SharePoint List to a given SPFx Form Customizer based on provided GUID. 
+- **Set Form Customizer** - Allows to update the New, Edit or View form of any SharePoint List to a given SPFx Form Customizer based on provided GUID. 
 
 ![Set Form Customizer](./docs/assets/images-vscode/set-form-customizer-action.png)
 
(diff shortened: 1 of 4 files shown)


Saurabh7019:
We have three 'Validate...' actions and I think the naming is confusing for all three. 
I know you are asking about the action name for 'Validate local setup for current project', but how about we make the other actions also clearer?

Here's what I propose:

1. **`Validate local setup`** - This is visible only when no SPFx project is open but 'local' in the name may be confusing as it does not indicate that this is a system/toolchain check. Rename it to **`Validate Global SPFx Environment`**
2. **`Validate project correctness`** - Fine as-is, but maybe we can add SPFx for consistency? Rename it to **`Validate SPFx Project Correctness`**
3. **`Validate local setup for current project`** - This should clearly indicate that it validates environment compatibility against the project's required version. Rename it to **`Validate SPFx Project Environment Compatibility`**

Adam-it:
>  for example - **Validate Global Environment** .. even within the SPFx Toolkit context, this doesn't make complete sense. Some may think, Validate global environment for what? 

That's why I think our action name is 
'validate local setup' 😉

Maybe to make it very specific it should be named 'validate node & global dependencies for SPFx dev' 🤔.

Names are important but I think it is important to find something that is short and to some extend quite clear to a person who works in this tech. For others we always have documentation and walkthroughs. 

Looking at the equivalent in CLI the command is 'spfx doctor'. IMO you either just know what it does or you need to use the --help option to figure it out 😁

Adam-it:
We just had a new minor release. We should rebase this branch before we proceed

Saurabh7019:
> maybe lets leave this topic for an upcoming maintianers call?

good idea :D @nicodecleyre might need to cast the deciding vote

nicodecleyre:
> > maybe lets leave this topic for an upcoming maintianers call?
> 
> good idea :D @nicodecleyre might need to cast the deciding vote

I can't handle that much stress!! 😄 

Ok for me to discuss it in our next call, but let's keep it simple.. If you ask me, i would go for:
1. Validate local setup 
2. Validate project correctness
3. Validate local setup for current project

PR #744, opened 2026-03-22
Adds prepare pre release flow. Closes #695

### LINKED ISSUE #695: Prepare pre-release workflow
Currently when we want to perform a pre-release we need to manually pull latest dev and manually bump the patch version in package.json and shrinkwrap and commit it to the repo.

If we are in a hurry it is almost impossible to do that over the phone, also mistakes happen and it's normal, if we would automate this we could prevent mistakes in future.

The idea is to add a new workflow that would bump the patch version for us on dev branch and would open a PR for us that targets dev branch and with title like "pre-release vx.y.z"
That way we may just quickly recheck on the PR if the modified files are correct, and merge like any other PR. 
After that we may just start the pre-release workflow as we do now.

That way we would reduce the amount of manual code we need to write to perform a pre-release and reduce the mistakes that may happen along the way.

Instead the pre-release flow would be just:

1. Run the prepare pre-release flow that does the bump and opens a PR
2. Quick check on the PR and merge
3. Run the pre-release flow that does the publish on the marketplace

Adam-it:
## 🎯 Aim

The aim of this PR is to introduce a new GitHub workflow which will prepare dev branch for running pre-release. 
That way if we will want to perform a pre-release from a mobile it will be as simple as:

1. Run the pre-release workflow
2. This will open a PR that targets dev branch with updated patch version in package.json and npm-shrinkwrap.json. We should review it, to double-check, and merge as any other PR
3. Run pre-release workflow

That's it. No manual local checkout and updating versions on our own.

## 📷 Result

I don't have any images to show, but I already tested it on my fork. 

- You may check the flow run here: https://github.com/Adam-it/vscode-viva/actions/runs/23413414115
- And this was the result PR: https://github.com/Adam-it/vscode-viva/pull/3

## ✅ What was done

- [X] Added a PS script that bumps the patch version in package.json
- [X] Updated package.json and npm-shrinkwrap.json formatting to 2 spaces - this change is a bit tricky. Currently, in package.json and npm-shrinkwrap.json, a single tab (4 spaces) is used for code formatting. Unfortunately, the PowerShell 7+ Convert-toJson method does not allow to set this and defaults to 2 spaces for formatting. So in order to have a clean PR with only changed version numbers I had to aling this formatting before hand to make it ready for first flow run
- [X] Added a prepare pre-release flow

## 🔗 Related issue

Closes: #695


Adam-it:
I need to rebase against dev

Saurabh7019 (approved):

Saurabh7019:
there were a couple of issues, the main one was Node.js 20 deprecation warning and outdated v4 actions. fixed those and merged. After merging I found a path issue in the script and fixed the same using a hotfix PR.

PR #755, opened 2026-04-17
Update actions used in GH workflows to latest. Closes #742

### LINKED ISSUE #742: Update actions used in GH workflows to latest
Currently, our workflows show the following warning

Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, actions/setup-node@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

We should recheck them and update the actions to the latest if possible

Saurabh7019:
## 🎯 Aim

> This PR fixes the Node.js 20 deprecation warnings by updating all workflow files to use the latest action versions that support Node.js 24. peter-evans/create-pull-request@v5 isn't updated to support Node 24 yet so replace that with native GitHub CLI commands.

## 📷 Result

> image

## ✅ What was done

- [X] Updated all workflows to support Node.js 24

## 🔗 Related issue

Closes: #742

### DIFF
--- a/.github/workflows/update-samples.yml
+++ b/.github/workflows/update-samples.yml
@@ -61,14 +61,26 @@ jobs:
           Remove-Item sample-stats.txt
         shell: pwsh
 
+      - name: Commit and push changes
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git checkout -b update-sample-data
+          git add -A
+          git diff --quiet && git diff --staged --quiet || git commit -m "Updates sample data"
+          git push origin update-sample-data --force
+        shell: pwsh
+
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@v5
-        with:
-          commit-message: Updates sample data
-          branch: update-sample-data
-          title: Sample data update
-          body: |
-            Automated check and update of SPFx extensions and webparts samples data and aces sample data.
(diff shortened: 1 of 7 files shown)


.github/workflows/update-samples.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   update:
-    if: github.repository_owner == 'pnp'
+    # if: github.repository_owner == 'pnp'
```
Adam-it: we need this
```suggestion
    if: github.repository_owner == 'pnp'
```

Adam-it (commented):
Only one small update

Adam-it (approved):
LGTM 🔥

.github/workflows/update-samples.yml
Saurabh7019: oh, I was testing it locally and forgot about it.

PR #759, opened 2026-04-20
Adds shortcut for the extension. Closes #732

### LINKED ISSUE #732: 💡 [Feature]: Shortcut to move to SPFx Toolkit view
### 🎯 Aim of the feature

in VS Code you may have shortcuts to move to views like Ctrl + Shift + E will move to the explorer view, or Ctrl + Shift + G, G moves to git view. 

I Wonder if we could have the same for SPFx Toolkit. When doing SPFx dev I am constantly jumping between the explorer and SPFx Toolkit view to reach out actions. I dont use command prompt that much

_No response_

_No response_

Adam-it:
## 🎯 Aim

Adds a shortcut that will allow you to quickly jump to the extension activity view

## 📷 Result

{3E4419CE-75D1-486D-A7BF-DA9B70CF56EA}

## 🔗 Related issue

Closes: #732

### DIFF
--- a/package.json
+++ b/package.json
@@ -961,6 +961,14 @@
 				}
 			]
 		},
+		"keybindings": [
+      {
+        "command": "workbench.view.extension.pnp-view",
+        "key": "ctrl+alt+s",
+        "mac": "cmd+alt+s",
+        "when": "true"
+      }
+    ],
 		"views": {
 			"pnp-view": [
 				{


package.json
```diff
@@ -961,6 +961,14 @@
 				}
 			]
 		},
+		"keybindings": [
+      {
+        "command": "workbench.view.extension.pnp-view",
+        "key": "ctrl+alt+s",
+        "mac": "cmd+alt+s",
```
Saurabh7019, line 968: @Adam-it , as per the issue spec, this should follow the standard pattern (Ctrl + Shift + <key>) used for views (e.g., Explorer, Search).

Any reason Ctrl + Alt + S was chosen? If not, can we switch it to Ctrl + Shift + S for consistency?

Adam-it: nope, `Ctrl + Shift + S` is used in VS code as shortcut to save all modified files

Saurabh7019 (approved):

PR #800, opened 2026-05-26
Removes node cache from docs workflow. Closes #791

### LINKED ISSUE #791: Recheck our GitHub workflows and confirm or remove using node cache

Adam-it:
## 🎯 Aim

The aim of this PR is to remove the use of node cache in our workflows as they do not prove safe. Currently we use it only in our release docs flow. Lets remove it and run `npm install` every time, this flow does not need to be fast, but safe

##  🔗 Related issue

Closes #791

### DIFF
--- a/.github/workflows/release-docs.yml
+++ b/.github/workflows/release-docs.yml
@@ -23,16 +23,7 @@ jobs:
         with:
           node-version: 22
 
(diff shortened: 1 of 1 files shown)


Adam-it:
closing this PR as I moved this change to be part of #806

PR #852, opened 2026-07-29
Adds language model tool that creates new spfx projects. Closes #707, #705

### LINKED ISSUE #705: Drop @spfx GitHub Copilot Chat participant
# 🎯 Aim

Copilot chat participants were the way of extending Copilot, like 1 year ago, when ask mode was the only mode AI tools had. 
Now everything is about agents and the way to extend AI apps like Copilot is to provide them with skills, tools, prompts, agents. This is what we already started in SPFx Toolkit as well by creating LM Tools. 

The aim is to remove/drop the spfx GitHub Copilot Chat participant and refactor all current chat commands to new dedicated LM Tools.

This issue will be used in order to track the progress on this work:

- [x] #706
- [ ] #707
- [x] #708

### LINKED ISSUE #707: Refactor the `/new` chat command to a dedicated SPFx new project LM tool
## 🎯 Aim

The aim is to remove all code related to `/new` chat command and instead create and SharePointFrameworkCreateNewProject Language Model Tool that will wrap the `yo` scaffolding command and execute it based on the user prompt in order to create a new project. This LM Tool for now should support only latest version of SPFx as the scaffolding form

## 🤔 Remark

Since in the next release od SPFx - 1.23 - the team will introduce a new way of creating SPFx projects by dropping `yo` support and using a dedicated SPFx CLI, I was wondering if maybe it would be more bneficail to wait for the 1.23 release and with this LM tool wrap the SPFx CLI create project command

Adam-it:
## 🎯 Aim

The aim of this PR is to depreacte the chat participant and remove the `/new` chat command and refactor it to a dedicated language model tool that will execute the best `yo` command to create and SPFx project for user prompt

## 📷 Result

image
image
image
image
image

## ✅ What was done

- [X] Removes the new chat participant command
- [X] Updates docs and readme
- [X] Adds a new langauge model tool that createst he best yo command for user prompt and executes it

## 🔗 Related issue

Closes #707, #705


Adam-it:
just noticed when doing different stuf that we should also remove 
https://github.com/pnp/vscode-viva/blob/62b56662cd272987c5e4f67a6e108a5b613ef9c0/src/panels/CommandPanel.ts#L365

and
https://github.com/pnp/vscode-viva/blob/62b56662cd272987c5e4f67a6e108a5b613ef9c0/src/panels/CommandPanel.ts#L376

and 
https://github.com/pnp/vscode-viva/blob/main/src/services/actions/CopilotActions.ts

with this change

Adam-it:
> just noticed when doing different stuf that we should also remove
> 
> https://github.com/pnp/vscode-viva/blob/62b56662cd272987c5e4f67a6e108a5b613ef9c0/src/panels/CommandPanel.ts#L365
> 
> and
> 
> https://github.com/pnp/vscode-viva/blob/62b56662cd272987c5e4f67a6e108a5b613ef9c0/src/panels/CommandPanel.ts#L376
> 
> and https://github.com/pnp/vscode-viva/blob/main/src/services/actions/CopilotActions.ts
> 
> with this change

done ✅

src/chat/tools/spfx/SpfxNew.ts
```diff
@@ -0,0 +1,110 @@
...
+                commandParts.push('--skip-install');
+            }
+
+            const yoCommand = commandParts.join(' ');
+
+            await Scaffolder.createProjectCopilot(yoCommand);
+
+            const response = `Created and executed the following SPFx command:\n\n\`\`\`\n${yoCommand}\n\`\`\``;
+
+            return new LanguageModelToolResult([new LanguageModelTextPart(response)]);
```
nicodecleyre, line 89: This works great, but not when you are in a workspace root on an existing SPFx solution where you already have a `package.json`. Then it produces a target folder containing only `.yo-rc.json`& `project.pnp`

Image

So... The generator logs `No change to package.json was detected. No package manager install will be executed.` and exits 0, because `createProjectCopilot` always runs `yo` with `workspace.workspaceFolders[0]` as current working directory, and the generator picks up that ambient `package.json` and treats the call as "nothing to do" instead of scaffolding a fresh, independent solution in the target subfolder.

Scaffolding into a workspace root that is not itself an existing SPFx project works fine every time.

src/chat/tools/spfx/SpfxNew.ts
```diff
@@ -0,0 +1,110 @@
...
+            }
+
+            if (params.skipInstall) {
+                commandParts.push('--skip-install');
+            }
+
+            const yoCommand = commandParts.join(' ');
+
+            await Scaffolder.createProjectCopilot(yoCommand);
+
```
nicodecleyre, line 86: `skipInstall` gets append if it's not already present unconditionally.

In the agent:

Image

In the output:

Image

nicodecleyre:
Great work so far @Adam-it 👏; Can you take a look at the comments i left? Thank you!

src/chat/tools/spfx/SpfxNew.ts
Adam-it: right. So you mean for a scenario in which you already have an SPFx project and would like to extend it with additional SPFx component right?

src/chat/tools/spfx/SpfxNew.ts
Adam-it: this is actually by design. We should scaffold a new project as fast as possible with skip install and reopen VS code in it and only then run npm install once VS Code is reopened in the newly created project. 
Do you think this follow is correct? or what do you think we should correct?

src/chat/tools/spfx/SpfxNew.ts
nicodecleyre: Yes, either it adds the right SPFx component, or it creates a new project in the existing project. But I would suggest the first 😄

src/chat/tools/spfx/SpfxNew.ts
nicodecleyre: You're right. Ok to scaffold a new project as fast as possible. But then again, if the user adds the skip install parameter, it will still install the packages when VS reopens. Doesn't that feels contradictory?

PR #853, opened 2026-07-29
Removes yarn package manager. Closes #834

### LINKED ISSUE #834: Remove Yarn Package Manager Support from SPFx Toolkit
Remove all yarn package manager support from the VS Code extension, leaving only npm and pnpm as supported package managers for SPFx development.

We need to do this because yarn is incompatible with the SPFx project upgrade command. yarn's CLI doesn't support setting package overrides, which is required by CLI for Microsoft 365's upgrade rules. Since yarn cannot be used for upgrades, supporting it for other operations creates an inconsistent user experience.

Adam-it:
## 🎯 Aim

The aim is to drop yarn package manager support

## 📷 Result

image

## ✅ What was done

- [X] Removed yarn from settings and all places where it was handled specifically

## 🔗 Related issue

Closes #834

### DIFF
--- a/package.json
+++ b/package.json
@@ -835,7 +835,7 @@
             "npm",
             "pnpm"
           ],
-          "description": "Choose your preferred Node package manager. Choose between `npm` and  `pnpm`."
+          "description": "Choose your preferred Node package manager. Choose between `npm` and `pnpm`."
         },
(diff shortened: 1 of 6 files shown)


Saurabh7019 (approved):
except for a minor typo, everything looks perfect.

PR #854, opened 2026-07-30
Adds saving multiple app regs for login. Closes #183

### LINKED ISSUE #183: 💡 [Feature]: Multiple accounts support
## 🎯 Aim of the feature

The aim is to add support for multiple accounts to the extension to manager your SPFx solution. 

~It should be possible to login to multiple accounts/tenants. and then using the account tab switch between the accounts and the actions done should respect the currently active account.~

instead lets develop what was proposed in the comment below

Adam-it:
## 🎯 Aim

The aim of this PR is to allow setting up multiple app regs for multiple tenants and save them in the extension settings. Then, when signing in, the user may pick the app reg by its friendly name from the list of saved app regs. It is also possible to delete one of the saved app regs. The PR does include a migration flow of the existing app reg the user might already have.

## 📷 Result

Saving app reg friendly name when creating a new app reg
image

The app reg name is visible in te view
image

new setting only editable directly in JSON settings mode
image
image

Deleting one of the saved app regs
image

## ✅ What was done

- [X] Added migration flow of existing app reg
- [X] Added saving app regs for sign in in VS Code extension settings with possibility to provide a friendly name
- [X] Added a possibility to remove an app reg
- [ ] TODO - I need to recheck the docs, most probably requires some screenshot updates BUT I will perform it once we go over the first review so we ensure I am on the right track here 😉

## 🔗 Related issue

Closes #183


src/services/dataType/AppRegistrations.ts
```diff
@@ -0,0 +1,107 @@
...
+   * @returns The last used app registration or undefined when there is none.
+   */
+  public static getLastUsed(context: ExtensionContext): AppRegistration | undefined {
+    const clientId = context.globalState.get<string>(LAST_USED_KEY);
+    if (!clientId) {
+      return undefined;
+    }
+
+    return AppRegistrations.getAll().find(item => item.clientId.toLowerCase() === clientId.toLowerCase());
+  }
```
nicodecleyre, line 66: `setLastUsed` stores only `clientId` and `getLastUsed` matches only on that. Elsewhere (`isSame`), a registration is unique on `clientId` & `tenantId`. I would suggest to store tenantId in `LAST_USED_KEY` (as a JSON object) and match on the pair, so that this is consistent with the rest of the class.

nicodecleyre (changes requested):
Great work so far! Can you take a look at the comment i left? And as you said, let's make the documentation before going for that final review 😉👍

Adam-it:
ok I resolved the comments and I will update  the docs in a bit

Adam-it:
All done. docs updated as well. @nicodecleyre ready for another round and lets get this shipped 🚢

src/providers/AuthProvider.ts
```diff
@@ -143,14 +155,134 @@ export class AuthProvider implements AuthenticationProvider, Disposable {
...
+    const quickPick = window.createQuickPick<AppRegistrationQuickPickItem>();
+    quickPick.title = SIGN_IN_TITLE;
+    quickPick.ignoreFocusOut = true;
+    quickPick.canSelectMany = false;
+    quickPick.items = getItems();
+
+    const lastUsed = AppRegistrations.getLastUsed(AuthProvider.context);
+    const lastUsedItem = lastUsed ? quickPick.items.find(item => item.registration?.clientId === lastUsed.clientId) : undefined;
+    if (lastUsedItem) {
+      quickPick.activeItems = [lastUsedItem];
```
nicodecleyre, line 234: `AppRegistrations.isSame` (and `getLastUsed` itself) treats `clientId` + `tenantId` as the compound identity key, but here both the "(last used)" label and the pre-selected `activeItems` entry are picked using clientId alone. Since a multi-tenant app registration can legitimately be saved twice with the same clientId but different tenantId signing in to tenant B would mark both the tenant-A and tenant-B entries as "(last used)" and could pre-select the wrong one. Compare against tenantId too, e.g. `lastUsed?.clientId === registration.clientId && lastUsed?.tenantId === registration.tenantId`.

nicodecleyre (changes requested):

Adam-it:
@nicodecleyre done.
lets get this shipped 🚢🚀

nicodecleyre:
> @nicodecleyre done. lets get this shipped 🚢🚀

Correct me if I'm wrong, but I don't see any changes per my last review? 😄

Adam-it:
> > @nicodecleyre done. lets get this shipped 🚢🚀
> 
> Correct me if I'm wrong, but I don't see any changes per my last review? 😄

sorry,  I had conflicts on push and did not notice. added the commits now. this is the one solving your comment 
https://github.com/pnp/vscode-viva/pull/854/changes/cf72f403fea63e841cc30411c6589828213f182d

lets move forward

nicodecleyre (approved):
Great work Adam, thank you for your contribution 💪

PR #861, opened 2026-08-01
Adds lock file linting to flows running with every PR. Closes #775

### LINKED ISSUE #775: Prevent npm lockfile injection
https://github.com/lirantal/npm-security-best-practices#4-prevent-npm-lockfile-injection

Adam-it:
## 🎯 Aim

The aim of this PR is to add a lint step for npm dependencies present in the lock file to every workflow which we run for PRs so that we try to shield ourselves from hacky updates done directly in the lock file in the PR that update the package hash to a vulnerable repo or package

## 📷 Result

{BEBE85F3-B9E1-4D28-B920-EC5A5F073613}

## ✅ What was done

- [X] Adds a new dev dependency lockfile-lint to use it in the pipeline
- [X] Updated the check tests workflow which is the only flow we run for every open PR and added step to lint lock file

## 🔗 Related issue

Closes #775


.github/workflows/check-tests.yml
```diff
@@ -35,6 +35,10 @@ jobs:
         env:
           npm_config_omit: optional
 
+      - name: Validate lockfile for SPFx Toolkit
+        run: .\node_modules\.bin\lockfile-lint --path npm-shrinkwrap.json --type npm --allowed-hosts npm --validate-https
```
nicodecleyre, line 39: Would you consider adding `--validate-integrity` and `--validate-package-names` to the lockfile-lint command? The current flags cover host and protocol validation, but the scenario described in the aim, a lockfile edit that swaps the package hash to a vulnerable repo or package, is specifically caught by `--validate-integrity`. `--validate-package-names` would additionally cover typosquatting or name mismatches between `package.json` and the lockfile.

package.json
```diff
@@ -1574,56 +1574,57 @@
...
     "@actions/core": "1.11.1",
-    "@estruyf/vscode": "1.2.0",
-    "@fluentui/react": "8.125.7",
-    "@tailwindcss/postcss": "4.3.3",
+    "@estruyf/vscode": "1.1.0",
+    "@fluentui/react": "8.122.16",
+    "@tailwindcss/postcss": "4.1.4",
     "@types/glob": "8.1.0",
     "@types/mocha": "10.0.10",
+    "@types/node-forge": "^1.3.11",
```
nicodecleyre, line 1582: Also noticed unrelated dependency changes (`@types/node-forge`. Removal of `@types/sinon`, `@types/vscode`, `@typescript-eslint-plugin`. Version swap on `rehype-raw`, `remark-gfm`, `use-debounce`). Were these intentional, or can they be split out of this PR since they're not needed for the lockfile-lint change?

nicodecleyre (changes requested):
Great work so far, can you check the comments i left? Thank you!

package.json
Adam-it: @nicodecleyre TBH those changes were just done 'automagically' after npm install. those are the same packages in the same versions just places in different order in the package.json. Not sure if the lint or npm install reoder this part 🙂.
I may revert to previous order but still nothing changed even now as the versions are the same

.github/workflows/check-tests.yml
Adam-it: good comment. lets do that 👍, will do that

Adam-it: added a dedicated config file with the settings

nicodecleyre:
@Adam-it, the test failed for this PR. It seems that 9 packages are failing because of an old sha1 hash. Can you take a look?

Adam-it:
> @Adam-it, the test failed for this PR. It seems that 9 packages are failing because of an old sha1 hash. Can you take a look?

ok added those old package to the exclude as well. We may move forward

nicodecleyre:
> > @Adam-it, the test failed for this PR. It seems that 9 packages are failing because of an old sha1 hash. Can you take a look?
> 
> ok added those old package to the exclude as well. We may move forward

Thank you! Just one more question: should the lockfile come before npm ci? so it blocks before downloading the packages? What do you think?

Adam-it:
> > > @Adam-it, the test failed for this PR. It seems that 9 packages are failing because of an old sha1 hash. Can you take a look?
> > 
> > 
> > ok added those old package to the exclude as well. We may move forward
> 
> Thank you! Just one more question: should the lockfile come before npm ci? so it blocks before downloading the packages? What do you think?

sure, good idea, will move it

Adam-it:
@nicodecleyre this is kinda tricky. 
On one hand it would be better to run the lockfile check before `npm ci` but then the `lockfile-lint` package would not exist and we use it to run the validation. If we would move it above `npm ci` then we would need to use `npx` in order to install and use `lockfile-lint` on the file, BUT using `npx` is a no go according to a different security improvement we already applied which states not to use `npx` 😵‍💫😵‍💫😵‍💫
So it is a bit of a closed circle  here 😁
@pnp/spfx-toolkit-maintainers what should we do with that?

PR #865, opened 2026-08-14
update update-package-version.ps1 to bump version in npm-shrinkwrap.json

### LINKED ISSUE #864: The Prepare Pre-Release workflow updates the version in package.json but not in npm-shrinkwrap.json, causing version inconsistency.
The workflow was changed from `npm install` to `npm ci` for security hardening. Unlike `npm install`, `npm ci` doesn't regenerate the lock file, so the `shrinkwrap` version stays outdated.

this issue is to update `update-package-version.ps1` to also update both version fields in `npm-shrinkwrap.json`

Saurabh7019:
## 🎯 Aim

This PR updates `update-package-version.ps1` to also update the version in `npm-shrinkwrap.json` when bumping the package version.

## ✅ What was done

- [X] updated `update-package-version.ps1` 

## 🔗 Related issue

Closes: #864

### DIFF
--- a/scripts/update-package-version.ps1
+++ b/scripts/update-package-version.ps1
@@ -9,3 +9,6 @@ $oldVersionParts = $oldVersion.Split(".")
 $newVersion = "$($oldVersionParts[0]).$($oldVersionParts[1]).$([int]$oldVersionParts[2] + 1)"
 $packageJson.version = $newVersion
 $packageJson | ConvertTo-Json -Depth 10 | Out-File "$PathToWorkspace\package.json" -Encoding utf8 -Force
+
+$shrinkwrapPath = "$PathToWorkspace\npm-shrinkwrap.json"
+(Get-Content $shrinkwrapPath -Raw).Replace("`"$oldVersion`"", "`"$newVersion`"") | Out-File $shrinkwrapPath -Encoding utf8 -Force
--- a/scripts/update-package-version.ps1
+++ b/scripts/update-package-version.ps1
@@ -11,4 +11,7 @@ $packageJson.version = $newVersion
 $packageJson | ConvertTo-Json -Depth 10 | Out-File "$PathToWorkspace\package.json" -Encoding utf8 -Force
 
 $shrinkwrapPath = "$PathToWorkspace\npm-shrinkwrap.json"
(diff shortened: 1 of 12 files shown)


scripts/update-package-version.ps1
```diff
@@ -8,4 +8,7 @@ $oldVersion = $packageJson.version
 $oldVersionParts = $oldVersion.Split(".")
 $newVersion = "$($oldVersionParts[0]).$($oldVersionParts[1]).$([int]$oldVersionParts[2] + 1)"
 $packageJson.version = $newVersion
-$packageJson | ConvertTo-Json -Depth 10 | Out-File "$PathToWorkspace\package.json" -Encoding utf8 -Force
\ No newline at end of file
+$packageJson | ConvertTo-Json -Depth 10 | Out-File "$PathToWorkspace\package.json" -Encoding utf8 -Force
+
+$shrinkwrapPath = "$PathToWorkspace\npm-shrinkwrap.json"
+(Get-Content $shrinkwrapPath -Raw).Replace("`"$oldVersion`"", "`"$newVersion`"") | Out-File $shrinkwrapPath -Encoding utf8 -Force
```
nicodecleyre: This does a raw string replace across the entire `npm-shrinkwrap.json` file, not a targeted update of the root package's version fields. If any transitive dependency ever happens to sit at the exact same version string as this project's current root version, this will silently rewrite that dependency's declared version too and corrupt the lockfile. It's safe today by coincidence, but that's not guaranteed going forward.

I would suggest replacing this with letting npm itself resync the lockfile, which can't touch the wrong entry:

`npm install --package-lock-only --ignore-scripts`

nicodecleyre (changes requested):
Great work Saurabh 👏. Can you take a look at the comment i left?

scripts/update-package-version.ps1
Saurabh7019: one more attempt.. I have now replaced the raw string replacement with structured JSON update.. 

intentionally avoiding `npm install` because we had moved away from it in favor of `npm ci` to reduce supply chain risk.

nicodecleyre (approved):
Great work @Saurabh7019! Thank you for your contribution! 💪

PR #881, opened 2026-09-09
Updates CI/CD action to support multiple SPFx versions. Closes #682 #523 #514

### LINKED ISSUE #514: Support Multiple SPFx versions
## 🎯 Aim

Currently SPFx Toolkit usually supports latest version of SPFx. Although it may work with older versions of SPFx than latest, usually many features of this Toolkit default to this. So the aim is to rebuild this behavior and try to add support to more (best all) SPFx versions possible.

## ✅ The Plan 

In order to allow to support all SPFx version possible we need to rethink/refactor the following functionalities:

- [x]  #515 - here we validate node and global packages if they are up to date with latest version of SPFx and we have our own custom logic around that. What I was thinking is to refactor it so that first we would run the [`m365 spfx doctor`](https://pnp.github.io/cli-microsoft365/cmd/spfx/spfx-doctor) command. This command checks the globally installed SPFx generator and when it is found it validates the version of Node, gulp-cli, and yo if they are compatible with the SPFx generator. This is exactly what we need! aaaand on top of that if the command fails (so it will not find any SPFx generator package installed) we could then run our current logic which means that we will validate to the currently latest version of SPFx.

- [x] #516 - currently here we always install dependencies for latest version of SPFx. But what if we reuse [this compatibility matrix](https://learn.microsoft.com/en-us/sharepoint/dev/spfx/compatibility#spfx-development-environment-compatibility) (I mean store a copy of it in our code base)/ [In CLI we do it this way](https://github.com/pnp/cli-microsoft365/blob/04e20e9dc7f70ab7c461e3365ead91aada490e6c/src/m365/spfx/commands/spfx-doctor.ts#L73-L644), and present a picker (default option, so when user will just hit enter, will be always latest version of SPFx) with all SPFx versions and install the global dependencies for the selected version of SPFx? Also Node! In SPFx Toolkit we already have a setting in which dev may set if they are using NVM or NVS or none

Image

if the dev set:
   - none - then we may only check if the currently installed node version is compatible with SPFx version selected. If it is selected we proceed with installing dependencies, if it is not we fail with error message and link to install node, or NVM or NVS (like we do now)
   - NVM or NVS - then we check if the currently active node version is compatible if yes we proceed and if not we present a warning informing the dev that this node version is not correct and which version of node should be active/installed with 2 actions:
    - abort and the dev needs to do the switch on his/hers own, 
    - install and in this case we run either NVM or NVS install commands and then the switch/use commands to set the correct version of node.

- [x] #667 - for now we agreed that the scaffolding options will support only the latest version of SPFx. We will need a follow up and a separate issue to implement the below:

    - **'Create a new project'** - Currently this form (and the add new component form) is just a static html added as webview in the product. What we would need to do is build the UI dynamically based on a JSON definition we would keep locally and we should have a JSON definition for every major SPFx version 

    - **@spfx /new** - the chat participant should leverage the JSON definitions from the point above and based on this figure out the correct yo command that should be run to scaffold an SPFx solution. We need to more clarify that 🤔

- [X] **'Validate project correctness'** - here we do not need to change anything 🚀🚀🚀🚀. Under the hood we use `m365 spfx project doctor` which determines the SPFx project version from .yo-rc.json file and based on this version it validates the project npm packages for that version. Tested it couple of times and seems we are done here ✅

- [x]  #517 - So what we could do is first we could run and check if there is any SPFx-generator installed and get its version. If there isn't then we could do what we do now which is providing guidance to upgrade to latest version of SPFx. But if we will find some SPFx-generator we could run the `m365 spfx project upgrade` with the `--toVersion` option and specify the version we found in the globally installed SPFx-generator.

- [x]  #523 - currently this feature will create a CI/CD workflow/pipeline which is aligned with latest version of SPFx. In both GitHub workflow and Azure DevOps pipelines there is a variable in which we specify the node version and currently it is always aligned with the node version valid for latest version of SPFx. We may not do much here at the moment as this is how CLI command but we could open a issue in CLI repo to change this behavior. Best if the CLI command used under the hood would check the version of SPFx in the currently used SPFx project for that command and then based on the compatibility matrix pick the correlated Node.js version. Currently this is on hold until the related issue in CLI repo will be developed and merged
  - [x] Related issue: https://github.com/pnp/cli-microsoft365/issues/6717
    - [x] PR opened for the related issue: https://github.com/pnp/cli-microsoft365/pull/6766

Also I have an idea for an additional action visible in the context of SPFx project
- [x] #518  - we may think of better name but what I was thinking is to allow an action that will run the `m365 spfx doctor` command in the context of the currently opened SPFx project in VS Code. In this case the CLI command check the SPFx version that is used in the project based on the .yo-rc.json file and validates if the globally installed dependencies like gulp-cli and Node version are correct for that version of SPFx 🤯. I think it is an awesome feature for folks that will open some older SPFx project and would like to quickly check if they are using correct version of Node.js and gulp-cli for that project.

### LINKED ISSUE #523: 💡 [Feature]: Update Scaffold CI/CD Workflow action to support multiple SPFx versions
### 🎯 Aim of the feature

Currently this feature will create a CI/CD workflow/pipeline which is aligned with latest version of SPFx. In both GitHub workflow and Azure DevOps pipelines there is a variable in which we specify the node version and currently it is always aligned with the node version valid for latest version of SPFx. 
I created a PR to update this in the CLI for Microsoft 365 repo https://github.com/pnp/cli-microsoft365/pull/6766
When it's merged, we should align the CJS version of CLI which is packaged with this extension so that we add this support also to SPFx Toolkit

_No response_

_No response_

### LINKED ISSUE #682: 🐞 Bug report:  CI-CD action generates workflows that still use gulp for SPFx 1.22.x projects
### ⭐ Priority

(Low)☹️ Something is a little off

### 📝 Describe the bug

The CI-CD action generates workflows that still use `gulp` for building and packaging when running in SPFx 1.22 projects. This is a mistake as from that version of SPFx we should use heft

### 👣 Steps To Reproduce

simply run the ci-cd action for an SPFx 1.22.x project

### 📜 Expected behavior

should generate workflow/pipeline that uses heft not gulp

### 📷 Screenshots

_No response_

### ❓SharePoint Framework Toolkit version

4.16.0

### ❓Node.js version

_No response_

### 🤔 Additional context

This bug should be fixed on the CLI for Microsoft 365 side first

Adam-it:
## 🎯 Aim

The aim of this PR is to upgrade CLI for the Microsoft 365 SPFx Toolkit package to latest in which I aligned all CI/CD command fixes. Now it will allow generating a pipeline/workflow for any SPFx version also using heft (npm run build) for newer versions of SPFx and gulp for older versions.

## 🔗 Related issue

Closes #682 #523 #514

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@grconrad/vscode-extension-feedback": "1.0.0",
-        "@pnp/cli-microsoft365-spfx-toolkit": "1.13.0",
+        "@pnp/cli-microsoft365-spfx-toolkit": "1.14.0",
         "node-forge": "1.4.0",
         "react-markdown": "10.1.0",
         "rehype-raw": "7.0.0",
@@ -1544,9 +1544,9 @@
       }
     },
     "node_modules/@pnp/cli-microsoft365-spfx-toolkit": {
-      "version": "1.13.0",
-      "resolved": "https://registry.npmjs.org/@pnp/cli-microsoft365-spfx-toolkit/-/cli-microsoft365-spfx-toolkit-1.13.0.tgz",
-      "integrity": "sha512-GuROpe+3BTPAnIKWToXIWwzMhmfNVlN9n3noWR7ksBgDEDNAqaE82mJyxt4OrYsGC4/fVD6bYMdVJGrPw1W95Q==",
+      "version": "1.14.0",
+      "resolved": "https://registry.npmjs.org/@pnp/cli-microsoft365-spfx-toolkit/-/cli-microsoft365-spfx-toolkit-1.14.0.tgz",
+      "integrity": "sha512-+ZLoiRMJym37vLqG529tZ5EdNBNLYTA6NZ+gujqQaHHeLjW5eP5nJHxNYa2xD0D+2LF+d4Uq2JxLHle7PxiwEw==",
       "hasShrinkwrap": true,
       "license": "MIT",
       "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1757,7 +1757,7 @@
   },
   "dependencies": {
     "@grconrad/vscode-extension-feedback": "1.0.0",
-    "@pnp/cli-microsoft365-spfx-toolkit": "1.13.0",
+    "@pnp/cli-microsoft365-spfx-toolkit": "1.14.0",
     "node-forge": "1.4.0",
     "react-markdown": "10.1.0",
     "remark-gfm": "4.0.1",


Adam-it:
@Saurabh7019 since I merged your previous PR maybe you wanna check this one. This was the one that was blocked by the previous one I wanted to open this week before I jump back to other repos next week

Saurabh7019 (approved):
Tested on 1.21.x and 1.22.x and it works well.

